PR #33, opened 2017-02-13
Attempt to get TacVests at correct drawing layer.

ProfoundDarkness:


NoImageAvailable:
Works like a charm when pawn is facing south or east/west. When facing north with a shell apparel equipped the shell is rendered over the vest whereas without shell the vest ends up overlapping the head.

NoImageAvailable:
Manually merged as part of https://github.com/NoImageAvailable/CombatExtended/commit/e2d8d2470777aa6ed3b14460d6d746042685962a

PR #51, opened 2017-02-22
Fixed VisibleAccessory apparel issues on downed/bedded pawns.

ProfoundDarkness:
Also added a tidbit of remarked code should there end up being a visible head accessory though I didn't want to try and refactor all the code without something to test it against.


ProfoundDarkness:
3rd time's the charm? sorry for the clutter...

PR #84, opened 2017-03-03
Fix moderate bloodloss frequently killing pawns.

Thorbane:
Large consciousness subtracting debuffs are very often fatal due to their combined effects with pain, drugs, brain injuries, etc.


NoImageAvailable:
How much of a problem is this? You'd need a 40% consciousness debuff in addition to moderate blood loss to die from this and where would you get that?

Thorbane:
Extreme pain does severely affect consciousness, I've seen just about everyone downed die at 40% bloodloss.

NoImageAvailable:
Looks like you found a magic number. Will look into the issue and have some adjustments in my next commit.

Issue #46, opened 2017-02-21
Implement pawns taking ammo and loadout items from pack animals.


Thorbane:
Would help pawns sent on caravans get their loadouts back after they place everything in their inventory on pack animals, and I would love to see raiders using pack animals in the future.

NoImageAvailable:
I haven't played around with the world travel system yet but it seems to me with the CR inventory system it would make more sense for pawns to keep their inventories while travelling.

The issue itself should be fixable relatively easily by adding a check for items in pack animal inventories to the loadout job.

RenaudF:
When pawns come back from travelling they will empty their inventory.

PR #90, opened 2017-03-03
Issue 46

ProfoundDarkness:
Attempted to address issue #46.

When a pawn is attempting to fulfill a loadout after it fails to find an item on the map it will then look in all pawns which are pack animals belonging to the pawn's faction or prisoners who are 'hosted' by the pawn's faction, if they have inventories that are not empty.

If it finds something in that search then it triggers a new job driver which has the pawn go to the pack animal/prisoner with the item in inventory, transfers the item from their inventory to 'my' inventory, and then if the pawn should equip the item, uses CE's code to equip the item from their inventory.

This can involve multiple trips if the item type the pawn wants is stacked in multiple inventories.  All other typical checks are done similar to before adding this.  Only minimal testing has been performed.  Made use of the mod "Nandonalt - Set-up Camp" to assist in testing.


ProfoundDarkness:
Oh yeah... unfortunately this was based on my previous messy branch... sorry.

PR #102, opened 2017-03-08
Attempted fix for Error Resolving Symbol.

ProfoundDarkness:
Didn't have time to run thorough tests.  In approximately 20 runs didn't get the error again.


NoImageAvailable:
Problem in this case is a few spots where WornApparel is used but not checked if its empty. Implemented those checks on my end, should fix this particular issue.

PR #108, opened 2017-03-27
Issue 64 completed.

ProfoundDarkness:
Started out with an idea to help someone else tackle this issue and stumbled on a way to just do the issue myself.

Basically, using Harmony and Infix, replace the Constant in GenRadial named RadialPatternCount as well as the -60/60 values in the two early loops inside SetupRadialPattern method.  The rest is handled by RimWorld logic.

Be warned that large radius values may be allowed but if the loops are still too small it won't look right.  Tried to constrain the range kept to keep things looking OK.

Also updated Harmony since I make use of features which received a bug fix.


ProfoundDarkness:
Issue number might be wrong, the one with turrets causing a red error text.

Issue #100, opened 2017-03-06
Add reload weapons job


Thorbane:
Colonist should reload all weapons in their inventory when they have time, to minimize micromanagement.

NoImageAvailable:
Should limit this somehow. I've had people in the past request the ability to unload ammo from their inventories (hence the "unload" command in the reload menu) and it would defeat the purpose if pawns just immediately loaded that same ammo back in.

Thorbane:
Hm, what about adding a timer to it, similar to keeping food in inventory for animal work?

ProfoundDarkness:
Unloading weapons the pawn has equipped/in inventory isn't a big deal to me but unloading weapons on the ground would be nice.  That way we can reclaim some of the ammo from raids when times are tough.

Valerate:
Is there and "End combat" event you could use to cause a reload? Or with hunting a "job well done" thing?

NoImageAvailable:
Shouldn't this be fixed as part of https://github.com/NoImageAvailable/CombatExtended/issues/114?

ProfoundDarkness:
With PR #140 merged that should be it for this issue.

Issue #34, opened 2017-02-13
Loadout suggestions


Thorbane:
The original "Nothing" loadout should not be alterable.
Add a generic option, ie. "carry x number of meals"
Interface with stock drug policy system's take to inventory feature
Add a copy loadout function, would save a lot of clicks in many situations.
Allow sorting of the assignment screen by loadout, for easier use.

RenaudF:
Give weapons in the loadout an auto-pick ammo option

ProfoundDarkness:
I've been pondering generics a bit in connection to other systems.

We could either do infixing/detours to engage with other systems (ie drug) or allow generics and a flag to handle the situation.  There are two parts to this.

Part 1 - Generics: I envision these (at least as a starting point) as being able to select a category much like how items are nested in categories for storage zone settings.  What shape that takes exactly is left for later.  Using the already available category system seems like the simplest method of handling things from a user perspective since they will probably already be familiar with item hierarchies.

Part 2 - Autofill loadout slot flag: The new flag would indicate if the pawn should try to fill the need or not.  This seems like the simplest path of implementation to make it so that pawns can pick up something (specific or generic) and the loadout system won't try to drop it later but nor will the loadout system try and fill that configuration.

Between these we could (By default when making a new loadout object) include two loadout slots, one for meals (category) (or even food) and another for drugs.  The new loadout items would have the flag indicating not to auto fill but that also means they wouldn't be auto-dropped.  Then we don't have to lift a finger to interact with the other systems.

NoImageAvailable:
Regarding categories, the first prototypes of the loadout system used the category menu instead of individual Things for assignment. It was very clunky and convoluted in addition to having some significant practical issues such as how to assign quantities. For those reasons it was discarded in favor of the current implementation.

ProfoundDarkness:
Items which, to my mind, have been dealt with as of commit #105 or before:
*The original "Nothing" loadout should not be alterable.
*Add a generic option, ie. "carry x number of meals" (Generics which are Lambda based)
*Interface with stock drug policy system's take to inventory feature (can use a Generic that ignores drugs so that the drug policy system's take to inventory won't be auto-dropped).
*Add a copy loadout function, would save a lot of clicks in many situations.

Things that haven't been done by that point:
*Allow sorting of the assignment screen by loadout, for easier use. (low hanging fruit after so much time on loadouts, adding to my list).
[...]

PR #114, opened 2017-04-04
Loadouts auto reload job+

ProfoundDarkness:
Small stuff:
* Loadouts are sorted for the user with default on top. (Accomplished by making Loadouts comparable).
* Ammo added via a gun's float menu "Add ammo for ..." sets a clip's worth of ammo for that gun and should fall back to defaultAmmoCount if the weapon doesn't use magazines (ie bows).
* Removed the assumption that making a new slot should be in defaultAmmoCount amounts so players can again define ammo 1 at a time if desired (per click).

Medium stuff:
* Added a new button to the Gear tab of pawns to generate a loadout from their inventory.  The generation tries to fit things that would be handled by certain types of generics into that bin but otherwise is ThingDef (Specific) based.  Also tries to add the common generics to the Loadout.  The idea to place it on the Gear tab comes from Karim.  The button doesn't show up if the pawn already has a loadout assigned or has nothing in their inventory.  It doesn't quite work the way I had hoped but it's about as close as I can get it until I understand more about RimWorld/Unity UI.

Bigger stuff:
* Addressed performance issues due to Generics, mostly in the Manage Loadouts UI.
+ Added an optimization to Ammo LoadoutGenericDef lambda - forgot to carry the optimization over from before AmmoLink.  (This helped a BUNCH.)
+ Modified ManageLoadout UI to only test 1 Generic against the items on the map per tick to spread the load out over time instead of checking all every time.  Still some hitching but not bad.
- It still checks every generic the first time the window is opened but rechecks are spread over ticks and no rechecks while paused (since no ticks then).  Switching to Generics tab causes about a 3/4ths second delay for me, reasonable for a one off IMO.
* Added an auto-reload mechanic via a new JobGiver which has a slightly lower priority than the high priority JobGiver_UpdateLoadout (9.2 vs UpdateLoadout's 9.3 and 3.0).  So pawns should reload after they have fetched nearby important stuff but before they try to top off their store/long distance travel.  Not sure if I'm using JobDriver_TryStartReload properly (might need to set some nulls for trying to resume an old job).

Bug Fixes: (Might have been some fixes I didn't note here)
* Fixed a bug due to my lack of understanding how ThingContainer.TryAdd() works and thus tried to put something in the Pawn's inventory AND drop it on the ground (bad).
* Fixed a bug introduced when the default Loadout became non-editable (basically an NRE when clicking the top most edit button on Assignment tab).
* Fixed an error from before I got into this code, though I did expose it.  When I started on this code base the Remove button in Manage Loadouts was removing items directly from the list instead of using LoadoutManager.RemoveLoadout().  While this did remove Loadouts it didn't unassign pawns who had that Loadout (also would have skipped any other work RemoveLoadout() might have done).  The bug was due to using an IEnumerable to alter the dictionary (and clauses of the IEnumerable), thus invalidating the IEnumerable.  Fix was switching over to a List so as to disconnect the Enumeration from the Dictionary.


ProfoundDarkness:
This might address issue #100.

It does address issue #34, at least the ones I think I can do something about.

ProfoundDarkness:
Regarding CompAmmoUser.TryStartReload(), I tried to do that to begin with and got seriously lost.  Moving the Mote Thrower code into the Reload JobDriver seemed to simplify the logic some.  I'll take another look.

Completely forgot about the ammo cache in CompInventory, will do.

NoImageAvailable:
Currently TryStartReload does these things:

1. Sanity check for gun magazine
2. Sanity check for wielder/turret existing
3. Sets turret to reload state, calculates weapon/turret position
4. Dumps currently loaded ammo into inventory/on ground (if unloading it stops here)
5. Throws a reloading text mote
6. Stores current job for later reassignment
7. Issues reload job

This commit already moves 3 and 5 to the JobDriver. We create a static utility method returning a Toil (something like vanilla's Toils_Tend class) that is a simple Wait-Toil with duration=reload time and performs 3-5 as init action, then adapt the JobDrivers for Reload, Hunt, WaitCombat and OperateTurret to include something like if(ammo = 0) yield return ReloadToil it would allow smoother integration into those Jobs (pawns would reload turrets even before firing for instance) and also reduce TryStartReload to only sanity checks and issuing the reload job.

ProfoundDarkness:
BTW: Pawns should already be reloading guns/turrets before warmup/firing automatically (empty ammo) now. (or I goofed something up.)

ProfoundDarkness:
OK... So broke up TryStartReload() into pieces that can be called separately.

TryUnload() which pretty much does the same as TryStartReload(true).  Calls to TryStartReload(true) are now TryUnload().

TryGetJob(bool) which tries to create the job to perform a reload, without actually applying it to the pawn. The argument indicates if it's a playerForced or not.

Both CompAmmoUser.TryStartReload() and JobGiver_CheckReload use CompAmmoUser.TryUnload() and .TryGetJob()

Added a new Method to CompInventory which functions similar to container.TotalStackCountOfDef() but operates on the CompInventory cache of Ammo Things.

JobGiver_CheckReload now uses that new method as well as the cache of Ranged Weapons instead of enumerating the pawn's inventory.

ProfoundDarkness:
I think I've addressed the requested changes.  If not let me know.

Regarding the idea of incorporating reload toils into the Hunting and other combat JobDrivers... while that is do-able (love me some Harmony) I'd rather not make a big break in the existing system right near a potential release when my free time is hard to pin down.  I can start working in that direction going forward but for now I tried to keep the existing functionality in place for code I wasn't messing with/reviewing.

Surprised just how much the original code for reloading guns on pawns pushed equipment only reloads, not sure if my making guns in inventory reload-able was apt.  Maybe add a penalty time for if the gun is in inventory and not equipment?

NoImageAvailable:
It's a bit late for that now but I would have suggested simply equipping the gun to reload, then switch back to the original gun after all inventory guns are reloaded. This way it would be more clear to the player what is going on as opposed to carried guns just inexplicably reloading themselves.

PR #121, opened 2017-04-18
Some tweaks to the project file now that I have VS...

ProfoundDarkness:
Moved some Debugger (build mode) specific things into the place that VS expects it and following some of the VS form for how to do things.

Also cleaned up the Xamarin options (those that Xamarin sees) as there was an outdated item there.

btw: Used TortoiseGit's diff feature to have some confidence that I didn't break something bigger in the project file.  That doesn't mean I didn't break something but I don't think I did.


ProfoundDarkness:
btw: Used TortoiseGit's diff feature to have some confidence that I didn't break something bigger in the project file.

PR #140, opened 2017-04-23
Modified motes and reportStrings for Reloading (includes turrets).

ProfoundDarkness:
Some of the remaining issues in #100.

Looking for feedback, in particular regarding turrets and the reportStrings in general.

-Removed "Reloaded!" mote (from CompAmmoUser).
-Added Keys (Translation strings) for motes, weapon source, and turret types.
-Modified Weapon and Turret Reload JobDrivers to use the new formatted motes.
-Added GetReport() override to generate a more detailed report of pawn activity.

Some light rearrangement and cleanup.

So reloading an equipped weapon shows the mote "Reloading Pistol!" and the reportString would say "Reloading equipped Pistol with <AmmoDef.LabelCap>".  Reloading an M240 turret shows the mote "Reloading turret M240" and the reportString "Reloading Manned Turret M240 with <AmmoDef.LabelCap>".

### DIFF
--- a/Defs/JobDefs/Jobs_CR.xml
+++ b/Defs/JobDefs/Jobs_CR.xml
@@ -6,7 +6,7 @@
 	<JobDef>
 		<defName>ReloadWeapon</defName>
 		<driverClass>CombatExtended.JobDriver_Reload</driverClass>
-		<reportString>Reloading weapon.</reportString>
+		<reportString>Reloading FlagSource TargetB with AmmoType.</reportString>
 	</JobDef>
 	
 	<JobDef>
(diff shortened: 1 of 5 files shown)


ProfoundDarkness:
Not really sure how to handle the whole non-combat situation.... If the pawn is drafted don't show mote?

NoImageAvailable:
Shouldn't it be non-capitals, i.e. "Reloading pistol." instead of "Reloading Pistol."?

ProfoundDarkness:
The report strings still have ammo with first word upper case but that seems to be in the field and not code.

Issue #142, opened 2017-04-23
Rescued space refugee desperately wanting equipment


Maeyanie:
Testing on v0.2.0-beta

My colony rescued a space refugee. When she was able to walk, she mined her way through the wall of her room, walked to my stockpile, and grabbed a knife.

Fortunately, Rimworld isn't a slasher movie, so she then went back to bed and lay down to rest instead of killing my colonists one by one. But still.

I'm not sure if this applies to anyone other than space refugees. I haven't seen it happen, but that doesn't mean it won't. I'll update this post if it does.

NoImageAvailable:
In the mod settings, do you have the auto-equip ammo option enabled? (On by default)

Maeyanie:
Yes, I do. All options are set to their defaults except for enabling right-click ammo selection.

dvalroth:
That's a bug that came up when Sky was maintaining it. Pawns were somehow getting raid AI assigned so they sapper their way to what they want rather than using doors like rational people.

Maeyanie:
The fact that they're trying to steal stuff in the first place is a bug, really. In vanilla they just are thankful to be able to walk freely again (unless they changed that since the last time I played vanilla, it has been a while).

I think I got something similar to happen with a released prisoner; he decided he wanted some clothes so picked them up on his way before leaving the map. Didn't bash through any walls, though, just picked up stuff that was on the ground outside a stockpile, so that might be legit.

If it is the same as the problems in Sky's version, it may also happen with sieging enemies and visitors if they spawn without a weapon, but I haven't had those happen while testing yet so can't report on it either way.

NoImageAvailable:
I know sky's additions included new AI behavior where raiders would pick up nearby weapons when out of ammo. IIRC it also included some behavior related to picking up apparel. My guess is its missing a sanity check somewhere so when you rescue a pod survivor it kicks in and tells him to find a weapon and ammo.

An easy way to check would be to start a new map, recreate the scenario then try again with the auto-equip ammo option turned off. If turning it off fixes it then we know for sure that's what's causing it.

ProfoundDarkness:
This was a bug I reported (or co-reported) in CR and someone else fixed... pretty sure CE branched from CR before that fix happened.

[...]

PR #157, opened 2017-04-29
Attempt to partially fix issue with visitors digging through walls...

ProfoundDarkness:
to get to a player's things.  Issue #142 

Mostly looked for logical errors, also cleaned up some code a little bit in the vicinity of the predicates I examined.  Noticed a TODO had been implemented already in the past and implemented another TODO to pickup ranged non-CE weapons.

Also fixed a bug (invalid cast) involving the new ammolink system and non-colonist pawns looking for guns to pickup.

Initial testing suggests that at least for visitors (non-hostile non-colonist pawns) this should be mostly fixed.

The overall concept for what this class should do could probably use examination.  Right now (at least for non-colonist pawns) the predicates and some other logic decides what a pawn should desire and then that pawn receives a GotoForce() job to get them there.  GotoForce() then has explicit instructions for pawns to dig if the pathfinder decides that is the apt path for them to follow.  Also looks like there is code to perform melee/use ranged weapon to break down walls but I don't think that code ever executes as there are two returns above that code, one if thing is not null and another if thing is null (thing being something in the pawn's way).

### DIFF
--- a/Source/CombatExtended/CombatExtended/Jobs/JobGiver_TakeAndEquip.cs
+++ b/Source/CombatExtended/CombatExtended/Jobs/JobGiver_TakeAndEquip.cs
@@ -261,7 +261,7 @@ namespace CombatExtended
                         Predicate<Thing> validatorWS = (Thing w) => w.def.IsWeapon
                         	&& w.MarketValue > 500 && pawn.CanReserve(w, 1)
                         	&& pawn.CanReach(w, PathEndMode.Touch, Danger.Deadly, true)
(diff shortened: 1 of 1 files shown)


NoImageAvailable:
I'm guessing the GotoForce() might be copied from vanilla somewhere (sapper-code or npc caravans digging their way out of player traps maybe?). Frankly I don't think it is needed at all because there are no circumstances where pawns should dig through walls to arm themselves, except (big maybe) prison breaks.

PR #161, opened 2017-05-02
Implemented an infix to change TendUtility to orderby BleedRate...

ProfoundDarkness:
ThenBy Severity.

Basically replaced part of/all of an original LinQ statement with a new one.  So less surgical than I'd like but born of my lack of knowledge and that the simpler approach was available.

Took some study since this was the first time I really looked at LinQ in IL.  If JUST BleedRate then would have been a simple prefix patch (helps to look at the code in a crummy debugger instead of the nicely patched ILSpy).


ProfoundDarkness:
Forgot to mention... the behavior is for any non-amputation injury the worst bleeder is first.  Amputations still get treatment at the end (would need to study the code further, might put amputations to first treatment blindly since they are usually the worst).  Among equal bleeders, the most severe is treated first.
Also in my small sample size medics seemed to start treating the worst bleeding patient first as well... this could have been coincidental but it was pretty consistent in my testing.

PR #191, opened 2017-05-31
Fragmentation overhaul

NoImageAvailable:
Added new gravity multiplier to fragments to reduce their max range.


NoImageAvailable:
Pushed to the wrong branch -_-

PR #232, opened 2017-06-07
Added danger check for pawns, when try to pick up ammo or weapon.

skyarkhangel:


ProfoundDarkness:
This looks like it's trying to fix one situation but also throws out a *bunch* of other tests.

I'm concerned that visiting *non-hostile* pawns might end up digging through bases (both open design but home marked and room partitioned) to get at stuff with this merged...  I also wonder about new bad behavior from hostile pawns.

Question: Is this primarily intended to replace the 'pawn.CanReach(w, PathEndMode.Touch, Danger.Deadly, true)' item?

NoImageAvailable:
Manually merged this with our homezone checks.

PR #247, opened 2017-06-12
Modified JobGiver_UpdateLoadout and JobDriver_TakeFromOther...

ProfoundDarkness:
In order to prevent an error which could occur when multiple pawns tried to take something from a pack animal...

Also modified JobGiver to check for reservation and reachability.

Modified JobDriver to make sure that the Thing the pawn wants to take is still in it's inventory and if the stack fell below what was desired take the rest of the stack.

Also added some additional fail conditions that seemed logical.  One might be redundant...

NOTES:
 1 - Need to test if my read of ReservationManager.CanReserve is accurate and if not either set reservations down to 1 (quick change in two sections of code) or add a more detailed check for any jobs reserving the animal are NOT JobDriver_TakeFromOther as a block. (I didn't see an easy way to go from pawn to list of jobs reserving that pawn without using Harmony to access internal Reservation data.)
 2 - The check if the thing is still in the takeFrom pawn's inventory isn't cheap however it's necessary as if a pawn took a full stack of something from the animal the Thing object was transferred rather than a new Thing created (that happens most of the time) meaning that it wouldn't be in takePawn's inventory anymore.  There might be a faster way (check if the container holding the thing is the takePawn's inventory rather than a search through takeFrom's inventory for Thing).

### DIFF
--- a/Source/CombatExtended/CombatExtended/Jobs/JobDriver_TakeFromOther.cs
+++ b/Source/CombatExtended/CombatExtended/Jobs/JobDriver_TakeFromOther.cs
@@ -20,18 +20,32 @@ namespace CombatExtended
 		private TargetIndex sourceInd = TargetIndex.B;
 		private TargetIndex flagInd = TargetIndex.C;
 		
+        /// <summary>
+        /// Property that converts TargetIndex.A into a Thing object.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


ProfoundDarkness:
Playing with an idea... don't merge yet.

PR #286, opened 2017-07-22
Added graphic swear motes

Killface1980:
+ peeing for hunkering pawns

Might need some refinement, but I think sometimes there is too much text on screen.


Killface1980:
Hi, have you tested it yet?

NoImageAvailable:
Sorry, this PR came in at a time when I was pretty busy with RL stuff and by the time I got back from my hiatus I had kinda forgotten about this. 

So after looking it over a couple of things:
1. Why use external resources? The game already has symbols for insults, using them would mean we don't need to rely on a third-party resource and muck around with licensing.
2. We removed the peeing from CE because I don't think its really appropriate or in good taste to have a feature like this in the mod, sorry.

NoImageAvailable:
Since this is quite outdated at this point and I have received no replies I'm closing this.

PR #287, opened 2017-07-22
Rimfire Patch

Valerate:
Rimfire Patch
Bolts Texture and Ammodef
Charge shotgun ammodef


NoImageAvailable:
There are a few issues with the patch:
1. ThingCategory Ammo6x24mmCharged_Shotgun currently serves no purpose as there are no ammo things assigned to it.
2. The damage values for the Charged Shotgun ammo seem way overpowered. If we compare the regular shotgun, one pellet of buckshot does 8 damage, about 2/3 of an assault rifle bullet. The charge shotgun on the other hand fires 9 full-power rifle bullets for the cost of one ammo.
3. In conjunction with 1. and 2. it seems kind of strange that the charge shotgun would run off of 6x24mm at all. Its like having a shotgun fire 5.56mm NATO. I think it would make much more sense to give it its own caliber.
4. The patch operations to remove various Rimfire research need to be contained in a PatchOperationSequence with a PatchOperationFindMod at the top, otherwise they'll produce log errors for anyone not running Rimfire.
5. Some guns, like the Taurus Judge still have market value defined in stats. CE uses relies on crafting costs and vanilla's automatic market value calculation, is there a reason why some of the guns should circumvent this?
6. The Large Rockets research is only intended for spacer level rocket launchers like the Doomsday and triple rocket launcher. For CE guns the Milkor and RPG-7 are simply part of the advanced guns research but if Rimfire has its own industrial level Small Rockets research it might make more sense to use that instead (dunno what the actual research requirements for that research are, if its also a spacer level tech just use advanced guns).
7. It seems a bit odd that all charge weapons would use 6x24mm. I already brought up the shotgun but we also have 6x18mm and 8x35mm which would be suitable for an SMG/sniper rifle respectively, so why not use those?

Valerate:
1. I barely know what I do, I\m sorry some unneeded stuff got a long
2. I have no idea how to balance them
3. I assume it was kind of a battery pack, and that all charge weapons used them (On older version of CR-Rimfire all chareg rifles used 6x24)
4. I have no idea what any of that means
5. Im sorry, that was a copy'paste mistake
6. Rimfire has its own small rockets research yeah, it was probably a better idea to keep it.
7. Again, I assume charge guns were enegy based weapons and that the 6x24ammo were just batteries

PR #303, opened 2017-08-09
Compatibilty with VFWE v2.2.2

XeoNovaDan:
Crossbow patched to work with CE, crossbow bolts added, improvised
pistol patched to be consistent with spreadsheet bar the fact it's now
stuffed, 'armour' to 'armor' for consistency.

### DIFF
--- a/Defs/Ammo/AmmoCategoryDefs/AmmoCategories_Neolithic.xml
+++ b/Defs/Ammo/AmmoCategoryDefs/AmmoCategories_Neolithic.xml
@@ -35,14 +35,14 @@
 		<defName>StoneSlingBullet</defName>
 		<label>stone sling bullet</label>
     <labelShort>stone</labelShort>
-		<description>A streamlined stone designed to fly far and hit hard. Not particularly effective against armour.</description>
+		<description>A streamlined stone designed to fly far and hit hard. Not particularly effective against armor.</description>
 	</CombatExtended.AmmoCategoryDef>
 
 	<CombatExtended.AmmoCategoryDef>
 		<defName>SteelSlingBullet</defName>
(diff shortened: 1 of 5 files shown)


Sakedo:
You'll also need to bring across your ammo tech level changes.

XeoNovaDan:
Ammo tech level changes? If you mean how it's structured in the directories, I probably should.

Sakedo:
You're using NeolithicAmmoBase as a parent in CrossbowBolts.xml:
<ThingDef Name="AmmoCrossbowBoltBase" ParentName="NeolithicAmmoBase" Abstract="True">

But that's undefined without your AmmoBase tech level changes. It's probably best to define just NeolithicAmmoBase and bring in the other changes in a separate pull when you think they're ready.

XeoNovaDan:
Eh, I'm unsure if this'll have any significant effect on gameplay anyhow. I think it'll be fine using the neolithic ammo base anyway.

Tribes are the closest things to medieval anyway in vanilla, seeing as they use medieval long bows the majority of the time instead of truly neolithic short bows.

Sakedo:
Yes, but NeolithicAmmoBase isn't defined in this pull request or the version of AmmoBases.xml that it's being merged into. it's in a commit on your local branch that isn't included with these changes.

XeoNovaDan:
I don't think it has to be defined in the PR because it's already defined in base CE, so it'll integrate anyway. It not being on the PR of course being because I haven't modified it.

Sakedo:
Oh right - this pull is going into the Development branch, where those changes have already landed.  My bad! Sorry to waste your time.

XeoNovaDan:
It's fine!

PR #320, opened 2017-09-12
More Ammo Types

AndrewTijua:
**Added more ammunition types.**

**Pistols**
-10mm Automatic (FMJ + AP + HP)
-.40 S&W (FMJ + AP + HP)
-.357 SIG (FMJ + ... yeah all of them have the same three types, you get the idea)
-.454 Casull

**Rifles**
-.30-06 Springfield (Because who doesn't love a good ping)
-.45-70 Government (because Rimworld is basically a western initially)

### DIFF
--- a/Defs/Ammo/Pistols/10mmAuto.xml
+++ b/Defs/Ammo/Pistols/10mmAuto.xml
@@ -1,0 +1,200 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<Defs>
+
+	<ThingCategoryDef>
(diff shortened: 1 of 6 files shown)


NoImageAvailable:
Why is this being added? Any mod using these?

AndrewTijua:
I'm working on one that does, not fully complete yet. Thought I should wait for the ammo to be included before releasing.

NoImageAvailable:
I'll have it in the next CE release. Let me know should you need it before that.

PR #329, opened 2017-10-25
Projectile arcs

Alicecomma:
PROJECTILES
- Height is now displayed visually on ALL projectiles. All projectiles also have a shadow beneath them to aid in this visualization. Projectiles fly in arcs and their rotation follows the arc.
- Projectiles shot by a "launcher != null && launcher is Pawn" are only displayed after
their first tick (so they don't spawn in the middle of the pawn).
- Projectiles collide with roofs and walls. This was tested thoroughly such that no projectiles could get through a walled, roofed building.

FRAGMENTS
- Fragments are aimed upwards/downwards based on whether a roof or floor is hit.

COOK-OFF
- Ammo cooks off at the middle of the tile, at height 0 c.

HITBOXES
- Plant and Pawn heights vary with growth and lifestage. Plant bounding boxes are now much larger but most of them don't stop projectiles due to a factor fillPercent often being 0f.
- Roofs now have hitboxes. Constructed roof adds a height of 0.1 c, natural roof adds 4 c (wallHeight*2), thick roof adds 4 c and thick natural roof (OverheadMountain) adds 8 c (thick * natural).
- BoundsInjector generates hitboxes based on Texture2D of Pawn/Plant. This can be expanded. See 5058b67
- This Texture2D method uses "blitting" which I have decided to limit to 64x64 px sprites and smaller (larger sizes are resized).

TUTOR
- The first time a player hovers over any crouched pawn, a Crouching learning thingy is added to the top right.

SUPPRESSION
- Suppression has a height limit equal to its radius.

CODE CHANGES
- ProjectileCE now uses mostly get methods (such as ShotLine) instead of parameters. ExactPosition and LastPos are Vector3 describing projectile position. ExactPosition can be set to stop the projectile dead in its tracks and treat is as landed.
- DrawPosition does never match ExactPosition / Position for projectiles.
- CompExplosive.Explode takes Vector3 for more precise positioning of an explosion.


Alicecomma:
TODO FOR BAREBONES RELEASE:
- Figure out what to do on roof intersection for both explosive and non-explosive projectiles
- Implement Texture2D-based Bounds calculations for Pawn and Plant

ALSO POSSIBLE:
- Suppressive fire over walls
- Dropshadows defined in XML per projectile
- Projectiles are displayed away from the visible gun (as to simulate bullets leaving a barrel) - currently solved by not displaying the first frame of the arc

Alicecomma:
89ad2c7 contains some fixes for #333

Issue #235, opened 2017-06-09
Reload report string has incorrect ammo


NoImageAvailable:
1. Have an out-of-ammo weapon with FMJ selected.
2. Have operator have AP or HP ammo in inventory.
3. Hit reload button with ammo still set to FMJ.

Weapon will load AP/HP ammo but report string says its loading FMJ.

Alicecomma:
The issue is due to the method _after_ the timer deciding what ammo to load/whether there is any ammo to load to begin with. This method should be split up into a part deciding whether anything could be loaded and what ammo to load (pre-timer) and a method actually performing the reloading (post-timer).

PR #349, opened 2017-11-14
Fix for #235

Alicecomma:
- CompAmmoUser:
- - private -> public TryFindAmmoInInventory
- JobDriver_Reload:
- - new field initAmmo
- - "AmmoType" in the reload string is replaced by initAmmo.LabelNoCount
- - FailOn(!tryFindAmmoInventory(out initAmmo))
- - LoadAmmo(initAmmo) as finish action


Alicecomma:
Additionally, reload will fail and unequip weapon if no usable ammo is present in inventory.

If ammo is removed during reload, the progress is kept and a different ammo is being reloaded.

PR #415, opened 2018-01-01
Power claw fix Slash -> Scratch (removed ToolCapacity Slash)

Alicecomma:
capacity Slash doesn't have a maneuver, but there's a maneuver called
Slash. It seems this was caused by a typo and prevented Power Claws from
being recognized as melee weapons


NoImageAvailable:
This would lead to power claws using the wrong damage def (scratch instead of slash). Correct solution here is to add the missing maneuver def for slashing attacks, especially since this bug is likely affecting more than just power claws.

PR #416, opened 2018-01-01
ArmorUtility.GetPenetrationValue revamp

Alicecomma:
- GetUsedTool(tools, dinfo) to get the tool specified in dinfo. Contains
a lot of error logging whenever it can't find the exact tool used.
- IsWeapon -> IsMeleeWeapon check (this previously caused errors on
meleeing with ranged weapons)
- More error logging for augments and natural tools, to make sure the
XML gets updated at some point such that the method is able to
differentiate between tools

### DIFF
--- a/Defs/Bodies/BodyPartGroups.xml
+++ b/Defs/Bodies/BodyPartGroups.xml
@@ -66,6 +66,11 @@
   </BodyPartGroupDef>
 
   <BodyPartGroupDef>
+    <defName>Muzzle</defName>
+    <label>muzzle</label>
+  </BodyPartGroupDef>
+
+  <BodyPartGroupDef>
     <defName>Barrels</defName>
     <label>barrels</label>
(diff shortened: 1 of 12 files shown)


Alicecomma:
[8:01 PM] Alistaire: @NoImageAvailable The errors added in latest PR are pretty in-your-face cause many tools can not be resolved from the damageinfo
[8:02 PM] Alistaire: A solution could be to add a lot more BodyPartDefs for weapons so they can be resolved
[8:02 PM] Alistaire: So there's two BodyPartDefs of barrel

Alicecomma:
I found a new way to sometimes distinguish some of them:

If the BodyPartDef is the same, look at capacities -> maneuvers with that capacity -> damagedef of the maneuver. If one of them fits the dinfo damagedef, it COULD be that one (since sometimes the damagedef is changed, e.g for critical hits and such).

PR #421, opened 2018-01-03
Animal won't knockdown the target if its size much larger.

skyarkhangel:


NoImageAvailable:
This should already be accounted for in the comparative crit chance. Creatures with high crit are less likely to be crit themselves.

PR #435, opened 2018-01-22
embrasures enhancement?

AndrewZo:
So, i redrawn an embrasures a little bit, and added blueprint textures. Psd included. Rimworld is always compressed custom textures somehow, it hard to match to original one in game, but i tried -.-


NoImageAvailable:
The blueprint is very nice, but the redrawn embrasure textures are darker than the originals, resulting in discoloration when connected to walls (see the screen).

AndrewZo:
yeah, i'll try to correct this. Don't know why, i used  extracted texture of wall from the game :(

PR #457, opened 2018-03-11
Added projectile check for null. For some reason ammo throw an error …

skyarkhangel:
…when ignited


NoImageAvailable:
When exactly does this issue occur?

skyarkhangel:
try to set fire on grenades, shells on the ground

PR #461, opened 2018-03-22
Fix for binoculars. Dummy bullet doesnt have projectilePropsCE.

skyarkhangel:


NoImageAvailable:
What exactly is this meant to fix?

skyarkhangel:
pawn that using binoculars throw an error. CE can't build shiftvecreport.

Issue #483, opened 2018-04-17
Thing combination


jinlan:
Allow multiple things to be combined into one. 
 - gun light
 - scope for guns
 - night vision
 - replaceable barrel
 - bipod, improve weapon handling, only works when behind a Thing with fillage
 - tripod(furniture), you can optionally install guns and ammos into it. Cheap, no fillage, extra weapon handling, requires pawn to operate
 - gun mounted grenade launcher 
 - red dot sight
 - laser sight, less advanced version of red dot sight, chance to notify enemy(make them come to attack you)
 - ballistic computer
 - holographic sight, best sight but quickly run out of power
 - etc, any other thing that can make Things more customizable.

This should not like "giddy-up" mod, which attach one to another. This should just make secondary Thing invisible and may or may not change primary Thing into another Thing, or maybe make the primary Thing be hibernated in case of gun mounted grenade launcher.

I can code, but I can't draw.

Btw, if at least half of this is implemented, tribe faction will need a buff, I suggest after we significantly enhanced power of guns, make it harder to obtain bullets.

Issue #476, opened 2018-04-12
Bullet should penetrate body, wall, etc when still has energy.


jinlan:
As titled. This is especially "should be" for AP ammo. This should be easily done by move the Destroy function call from ProjectileCE.Impact() to BulletCE.Impact(), and only Destroy bullet when damage reduced to 0.

Possible related work:
 - Change wall's HP calculation mechanism from "baseHP * material HP factor" to "baseHP with CE_BodyPartGroupDefOf.CoverByNaturalArmor", armor value should be some number transformed from material HP factor.
 - Should ammo be left inside Pawn's body, and requires a surgery to take it out?

PS: Just tried, Much more difficult than that.

Done. This will significantly buff all kinds of machine guns.
Energy lose formula:
Pawn= bodysize * 0.2
other=HP/40 * 0.2 ( every 40 HP equals to 1 bodysize , use real HP not max HP )
if final remaining penetrating > 0.2 , then penetrate, continue to fly.

PR #490, opened 2018-04-25
Tw

jinlan:
Bullet Penetration - implemented #476

Allow bullet to penetrate body, walls, etc.
When penetrating body, generated armor = ( MaxHP of all hit organs ) * 0.006.
When penetrating other things, generated armor = MaxHP * 0.006
If remaining AP > 0.15, then penetrate. 
When penetrated any thing other than a pawn, generate 50% extra suppression.
After penetration, bullet will fly in the same rotation,  current height, reduced speed.

------------------------------------------------------------------------------------
partially implemented #483. Implemented changeable barrel. Now we can have fn scar and p320.

Example:

    <changeBarrelTime>12</changeBarrelTime>
    <changeableBarrels>
        <li>
           <ammoSet>AmmoSet_545x39mmSoviet</ammoSet>
           <magazineSize>60</magazineSize>
       </li>
       <li>
           <ammoSet>AmmoSet_762x39mmSoviet</ammoSet>
           <magazineSize>40</magazineSize>
       </li>
    </changeableBarrels>

changeBarrelTime: How long time in seconds need to change barrel for this gun.
changeableBarrels: A list of changeable barrel.
ammoSet: Ammo set of this barrel
magazineSize: Magazine size of this barrel

Those markup must under the \<li Class="CombatExtended.CompProperties_AmmoUser"> element. It's actually "changeBarrelTime" property and "changeableBarrels" property of CombatExtended.CompProperties_AmmoUser class.

### DIFF
--- a/Defs/Ammo/Advanced/12GaugeCharged.xml
+++ b/Defs/Ammo/Advanced/12GaugeCharged.xml
@@ -104,6 +104,7 @@
       <armorPenetration>0.35</armorPenetration>
       <pelletCount>9</pelletCount>
       <spreadMult>17.8</spreadMult>
+      <allowPenetrateTrough>false</allowPenetrateTrough>
     </projectile>
   </ThingDef>
 
@@ -119,6 +120,7 @@
         </li>
       </secondaryDamage>
       <armorPenetration>0.6</armorPenetration>
+      <allowPenetrateTrough>false</allowPenetrateTrough>
     </projectile>
   </ThingDef>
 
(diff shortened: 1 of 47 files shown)


Alicecomma:
Where do colonists hide the spare barrels and how come changing a barrel changes the magazine size as well? To me it feels like a quick and dirty implementation of weapon modification, however I do say I like the way modification is abstracted in something as quick and accessible as it is.



For the bullets penetrating through targets, I do agree it's something worth looking into but the implementation as-is requires lots of reading and generally, NoImageAvailable nor me are in the mood to check this much code. I'm sorry! Your best bet would be to go on Discord and discuss the changes, e.g some sort of defence of why it should be added to the code base.

In particular, implementation of the penetration mechanics then making nearly every type of bullet non-penetrating is an interesting choice, especially considering you suggested in-depth calculation of whether penetration is achieved or not. This strongly indicates that these in-depth calculations were poorly designed (otherwise there would be no need to exclude nearly half of the ammo types from the functionality), which is a bad sign while reviewing the commit.

PR #525, opened 2018-10-07
Fixed some Method Calls

DoomOfMax97:
Armor Penetration needed. (Probably something else than 0 in some parts)
InsertIntoLog seems to be AssociateWithLog now.
NameStringShort seems to be Name.ToStringShort now.
Harmony fix.
Commented "DebugViewSettings.drawInterceptChecks" out.
Might have fixed the SmokepopBelt.
And more stuff, check the Commits for more info.

Hope this helps


Source/CombatExtended/CombatExtended/CritDefGenerator.cs
```diff
@@ -11,15 +11,15 @@ namespace CombatExtended
     public static class CritDefGenerator
     {
         public static IEnumerable<DamageDef> ImpliedCritDefs()
-        {
-            foreach(DamageDef current in from def in DefDatabase<DamageDef>.AllDefs where def.externalViolence select def)
+        {                                                                           //TODO: ExternalViolence for what?
```
NoImageAvailable: Crit defs are obsolete now that armor penetration can be passed via damage info. They only existed so sharp weapon crits could bypass armor. Should increase penetration value in Verb_MeleeAttackCE instead.

Source/CombatExtended/CombatExtended/Loadouts/Dialog_ManageLoadouts.cs
```diff
@@ -73,7 +73,9 @@ public Dialog_ManageLoadouts(Loadout loadout)
             SetSource(SourceSelection.Ranged);
             doCloseX = true;
             closeOnClickedOutside = true;
-            closeOnEscapeKey = true;
+            closeOnCancel = true;
```
NoImageAvailable: Does the vanilla outfit manager have corresponding flags set?

Source/CombatExtended/CombatExtended/Projectiles/BulletCE.cs
```diff
@@ -21,7 +21,9 @@ private void LogImpact(Thing hitThing, out BattleLogEntry_RangedImpact logEntry)
 					hitThing,
 					intendedTarget,
 					equipmentDef,
-					def);
+					def
+                    //CoverDef Missing!
```
NoImageAvailable: CE doesn't use cover in the sense vanilla does

Source/CombatExtended/CombatExtended/Projectiles/BulletCE.cs
```diff
@@ -47,10 +49,12 @@ protected override void Impact(Thing hitThing)
                 // launcher being the pawn equipping the weapon, not the weapon itself
                 int damageAmountBase = def.projectile.GetDamageAmount(CE_Utility.GetWeaponFromLauncher(launcher));
                 DamageDefExtensionCE damDefCE = def.projectile.damageDef.GetModExtension<DamageDefExtensionCE>() ?? new DamageDefExtensionCE();
+                float armorPenetration = def.projectile.GetArmorPenetration(CE_Utility.GetWeaponFromLauncher(launcher));
```
NoImageAvailable: This should be taken from projectilePropsCE, vanilla method will return bogus values.

Source/CombatExtended/CombatExtended/Projectiles/BulletCE.cs
```diff
@@ -47,10 +49,12 @@ protected override void Impact(Thing hitThing)
                 // launcher being the pawn equipping the weapon, not the weapon itself
                 int damageAmountBase = def.projectile.GetDamageAmount(CE_Utility.GetWeaponFromLauncher(launcher));
```
NoImageAvailable: Vanilla method, does this return correct values for projectiles?

Source/CombatExtended/CombatExtended/Projectiles/ProjectileCE.cs
```diff
@@ -432,10 +432,10 @@ private bool CheckForCollisionBetween()
                 return false;
             }
 
-            if (DebugViewSettings.drawInterceptChecks)
+            /*if (DebugViewSettings.drawInterceptChecks)
```
NoImageAvailable: Why are these commented out?

Source/CombatExtended/CombatExtended/Projectiles/ProjectileCE.cs
```diff
@@ -639,7 +639,7 @@ private void ApplySuppression(Pawn pawn)
                 && pawn.Faction != launcher?.Faction
                 && (shield == null || shield?.ShieldState == ShieldState.Resetting))
             {
-                suppressionAmount = def.projectile.damageAmountBase;
+                suppressionAmount = def.projectile.damageDef.defaultDamage;
```
NoImageAvailable: This must be actual projectile damage

Source/CombatExtended/CombatExtended/SecondaryDamage.cs
```diff
@@ -22,6 +22,7 @@ public DamageInfo GetDinfo(DamageInfo primaryDinfo)
         {
             var dinfo = new DamageInfo(def,
                             amount,
+                            0, //Armor Penetration
```
NoImageAvailable: This will break secondary damage, needs proper AP value

Source/CombatExtended/CombatExtended/Verbs/Verb_MeleeAttackCE.cs
```diff
@@ -250,7 +250,8 @@ private IEnumerable<DamageInfo> DamageInfosToApply(LocalTargetInfo target, bool
             Vector3 direction = (target.Thing.Position - base.CasterPawn.Position).ToVector3();
             Thing caster = this.caster;
             BodyPartHeight bodyRegion = GetBodyPartHeightFor(target);   // Add check for body height
-            DamageInfo mainDinfo = new DamageInfo(damDef, GenMath.RoundRandom(damAmount), -1f, caster, null, source);
+            DamageInfo mainDinfo = new DamageInfo(damDef, GenMath.RoundRandom(damAmount), damDef.defaultArmorPenetration, //Armor Penetration
```
NoImageAvailable: Must be actual weapon AP

Source/CombatExtended/CombatExtended/ArmorUtilityCE.cs
```diff
@@ -286,7 +286,7 @@ private static float GetPenetrationValue(DamageInfo dinfo)
                     if (instigatorPawn.def == dinfo.Weapon)
                     {
                         // meleeVerbs: all verbs considered "melee worthy"
-                        Verb availableVerb = instigatorPawn.meleeVerbs.TryGetMeleeVerb();
+                        Verb availableVerb = instigatorPawn.meleeVerbs.TryGetMeleeVerb(dinfo.IntendedTarget);
```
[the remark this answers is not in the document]
DoomOfMax97: Do I get the correct penetration info from the dinfo? (As in vanilla or CE)
If so, I could do a refactor.

Source/CombatExtended/CombatExtended/Projectiles/BulletCE.cs
DoomOfMax97: What do you suggest to pass as the argument?
The method needs a variable of the type CoverDef.

Source/CombatExtended/CombatExtended/Projectiles/BulletCE.cs
DoomOfMax97: I don't know, I only have a basic understanding of Rimworlds methods and classes.
Will try to check, I assume not, since they are defined in a custom config.

Source/CombatExtended/CombatExtended/Projectiles/BulletCE.cs
DoomOfMax97: Okay will use that variable.

Source/CombatExtended/CombatExtended/Projectiles/ProjectileCE.cs
DoomOfMax97: The value "drawInterceptChecks" of "DebugViewSettings" does not exist anymore,
and I was unable to find anything similar. Unless a custom debug setting/button is planned I don't know how you would trigger it ingame.

Source/CombatExtended/CombatExtended/Projectiles/ProjectileCE.cs
DoomOfMax97: How do I get the damage?

Source/CombatExtended/CombatExtended/SecondaryDamage.cs
DoomOfMax97: How do I get the value? Doesn't secondary damage ignore armor, or am I misremembering something?

Source/CombatExtended/CombatExtended/Verbs/Verb_MeleeAttackCE.cs
DoomOfMax97: How do I get the value?

Source/CombatExtended/CombatExtended/Loadouts/Dialog_ManageLoadouts.cs
DoomOfMax97: No Idea I just made an educated guess in most cases. In which class would that be?

DoomOfMax97:
Definitely double check my changes, I know how to code, but I don't know the Rimworld API nor the whole code and data structure of CE/Rimworld.

Source/CombatExtended/CombatExtended/Projectiles/BulletCE.cs
DoomOfMax97: Hmm seems like this was there before my changes, I could use projectilePropsCE instead.

DoomOfMax97: Seems like it increases the damage based on a multiplier.
Source:
https://github.com/josh-m/RW-Decompile/blob/master/Verse/ProjectileProperties.cs#L76

Source/CombatExtended/CombatExtended/ArmorUtilityCE.cs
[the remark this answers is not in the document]
DoomOfMax97: Commented the method out, and replaced calls accordingly, see commit for more info.

Source/CombatExtended/CombatExtended/CritDefGenerator.cs
DoomOfMax97: Sounds like optimization, but I will leave it as is for now, since it should at least run before we remove redundant code. (In my opinion at least)

DoomOfMax97:
Done with current iteration.

DoomOfMax97:
Should be almost done with a release, but need to doublecheck the code later again.

DoomOfMax97:
We will have to look at possible multithreading issues due to asynchronous event exceptions.

Source/CombatExtended/CombatExtended/ArmorUtilityCE.cs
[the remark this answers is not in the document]
NoImageAvailable: Yeah, idea is we don't need to infer penetration through obtuse hacks since we can now pass it directly via the damage info from projectile -> armor system

Source/CombatExtended/CombatExtended/Comps/HediffCompProperties_InfecterCE.cs
```diff
@@ -8,13 +8,8 @@
 
 namespace CombatExtended
 {
-    public class HediffCompProperties_InfecterCE : HediffCompProperties
+    public class HediffCompProperties_InfecterCE : HediffCompProperties_Infecter
```
NoImageAvailable: What's the rationale behind this change?

Source/CombatExtended/CombatExtended/Comps/HediffComp_InfecterCE.cs
```diff
@@ -11,22 +11,53 @@ namespace CombatExtended
     public class HediffComp_InfecterCE : HediffComp
     {
         private const float infectionInnerModifier = 3f;                            // Injuries to inner body parts will be this much more likely to get infected
-        private static readonly IntRange InfectionDelayHours = new IntRange(6, 12); // Infections will appear somewhere within this timeframe
+        //private static readonly IntRange InfectionDelayHours = new IntRange(6, 12); // Infections will appear somewhere within this timeframe
```
NoImageAvailable: What's with these changes?

Source/CombatExtended/CombatExtended/Loadouts/Dialog_ManageLoadouts.cs
NoImageAvailable:
```csharp
// RimWorld.Dialog_ManageOutfits
public Dialog_ManageOutfits(Outfit selectedOutfit)
{
	this.forcePause = true;
	this.doCloseX = true;
	this.doCloseButton = true;
	this.closeOnClickedOutside = true;
	this.absorbInputAroundWindow = true;
	if (Dialog_ManageOutfits.apparelGlobalFilter == null)
	{
		Dialog_ManageOutfits.apparelGlobalFilter = new ThingFilter();
		Dialog_ManageOutfits.apparelGlobalFilter.SetAllow(ThingCategoryDefOf.Apparel, true, null, null);
	}
	this.SelectedOutfit = selectedOutfit;
}
```
For consistency, should keep the relevant options the same, so no close on cancel/accept.

Source/CombatExtended/CombatExtended/Projectiles/BulletCE.cs
NoImageAvailable: Yeah, double-checked it and this is the correct method to use.

Source/CombatExtended/CombatExtended/Projectiles/ProjectileCE.cs
NoImageAvailable: It does correspond to that button. If it got moved to something else these calls should be moved accordingly. If it was removed then they need to move to the CE debug menu.

Source/CombatExtended/CombatExtended/SecondaryDamage.cs
```diff
@@ -22,6 +22,7 @@ public DamageInfo GetDinfo(DamageInfo primaryDinfo)
         {
             var dinfo = new DamageInfo(def,
                             amount,
+                            primaryDinfo.ArmorPenetrationInt, //Armor Penetration
```
NoImageAvailable: Don't have all the code in my head right now but IIRC this will introduce issues where primary damage can penetrate deeper than secondary and vice versa. Need to review DamageWorker code to make sure secondary damage is applied correctly.

NoImageAvailable: After reviewing it looks like the DamageWorker code is going to need some fundamental restructuring, including wrt to how secondary damage is handled.

Source/CombatExtended/CombatExtended/Verbs/Verb_MeleeAttackCE.cs
```diff
@@ -409,7 +415,8 @@ private void DoParry(Pawn defender, Thing parryThing, bool isRiposte = false)
                 if (parryThing is Apparel_Shield)
                 {
                     // Shield bash
-                    DamageInfo dinfo = new DamageInfo(DamageDefOf.Blunt, 6, -1, defender, null, parryThing.def);
+                    DamageInfo dinfo = new DamageInfo(DamageDefOf.Blunt, 6, 0, //Armor Penetration
```
NoImageAvailable: AP 0 would disable shield bash damage. Should either use default value or get it from shield (better solution but don't know off-hand how difficult to implement). Also I think we found a previous version bug here since I don't think shield AP is ever actually handled in B18.

Source/CombatExtended/Harmony/Harmony-SmokepopBelt.cs
```diff
@@ -26,8 +26,8 @@ public static bool Prefix(SmokepopBelt __instance, DamageInfo dinfo)
                 float statValue = __instance.GetStatValue(StatDefOf.SmokepopBeltRadius, true);
                 DamageDef smoke = DamageDefOf.Smoke;
                 Thing instigator = null;
-                ThingDef gas_Smoke = ThingDefOf.Gas_Smoke;
-                GenExplosion.DoExplosion(position, map, statValue, smoke, instigator, -1, null, null, null, gas_Smoke, 1f, 1, false, null, 0f, 1, 0f, false);
+                ThingDef gas_Smoke = ThingDefOf.Gas_Smoke;                             //TODO: postExplosionSpawnThingdDef? \|/ (gas_Smoke ?)
```
NoImageAvailable: This just mirrors vanilla smokepop, no additions necessary.

Source/CombatExtended/CombatExtended/Projectiles/BulletCE.cs
NoImageAvailable: Should be safe to null

Defs/Ammo/Advanced/12GaugeCharged.xml
```diff
@@ -22,7 +22,7 @@
 	
 	<!-- ==================== Ammo ========================== -->
 
-  <ThingDef Name="12GaugeChargedBase" ParentName="SmallAmmoBase" Abstract="True">
+  <ThingDef Class="CombatExtended.AmmoDef" Name="12GaugeChargedBase" ParentName="SmallAmmoBase" Abstract="True">
```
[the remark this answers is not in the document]
DoomOfMax97: For some reason, I get an error for every single Ammo about nonexistent (AmmoDef specific) variables (It's looking in ThingDef for them). Normally the Class part of baseAmmo should be abstracted throughout all its children, but for some reason or change, it does not. This is only a brute-force solution, I wrote in the commit how to revert this change anytime.

Source/CombatExtended/CombatExtended/Comps/HediffCompProperties_InfecterCE.cs
DoomOfMax97: HediffComp_InfecterCE was based on it and I thought it was just a copy from an old Rimworld version. Also this allowed me to fix other things for now.

Source/CombatExtended/CombatExtended/Comps/HediffComp_InfecterCE.cs
DoomOfMax97: There were some compilation errors in here, and I was unsure which part was custom and which was not. I was not able to fix it fast enough since you know your own code better than I do, I focused on other errors, and replaced it with the updated Rimworld code it was based on for now.

Source/CombatExtended/CombatExtended/Verbs/Verb_MeleeAttackCE.cs
DoomOfMax97: Looking at here, 0 seems to be the default value:
https://github.com/josh-m/RW-Decompile/blob/master/Verse/DamageInfo.cs#L172

Source/CombatExtended/Harmony/Harmony-SmokepopBelt.cs
DoomOfMax97: Okay removing TODO.

Source/CombatExtended/CombatExtended/Loadouts/Dialog_ManageLoadouts.cs
DoomOfMax97: Changed.

Source/CombatExtended/CombatExtended/Projectiles/BulletCE.cs
DoomOfMax97: Is null.

Source/CombatExtended/CombatExtended/Projectiles/ProjectileCE.cs
DoomOfMax97: Replaced it with Controller.settings.DebugDrawInterceptChecks, and created variables accordingly.

Source/CombatExtended/CombatExtended/Comps/HediffComp_InfecterCE.cs
DoomOfMax97: Oh my bad, sorry.
I will revert it, but the Method tendTick of HediffComp_TendDuration is missing,
what do you suggest to fix it?
[HediffComp_TendDuration compTended = parent.TryGetComp<HediffComp_TendDuration>();]

DoomOfMax97: tendTicksLeft would be available, but it would need a complete refactor of that part. Unless I am missing something of course.

DoomOfMax97: Also, I would suggest taking a look at
https://github.com/josh-m/RW-Decompile/blob/master/Verse/HediffComp_Infecter.cs
to make sure nothing important in terms of feature or balancing got added.
InfectionChanceFactorFromTendQualityCurve and
InfectionChanceFactorFromSeverityCurve for example.

NoImageAvailable: The issue here is we need to calculate the amount of time a hediff was untended. Since the exact tick it was tended is no longer available we'd need to either recreate this functionality or infer it from the existing variables. Easiest way to handle it would probably to add a counter class variable and increase it in CompPostTick if the hediff is untended.

Source/CombatExtended/CombatExtended/Verbs/Verb_MeleeAttackCE.cs
NoImageAvailable: Default value as in one value to use for all shields. DamageInfo defaults to 0 but the way CE armor works it'll just get completely deflected by anything.

Defs/Ammo/Advanced/12GaugeCharged.xml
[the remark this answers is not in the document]
DoomOfMax97: Also, as a side note, StuffEffectMultiplierArmor has replaced ArmorRating_Sharp and ArmorRating_Blunt, in the case of Apparel_SimpleHelmet, Apparel_AdvancedHelmet and possibly more. So it should be replaced in Items_Resource_Stuff.xml, for each material.

Source/CombatExtended/CombatExtended/Verbs/Verb_MeleeAttackCE.cs
DoomOfMax97: *Should* be fixed.

Source/CombatExtended/CombatExtended/Comps/HediffComp_InfecterCE.cs
DoomOfMax97: You know the Rimworld code better and can implement it faster, I would have to figure out what exactly tendTicksLeft does, without any documentation.

DoomOfMax97: Okay pushed a potential fix but I am not sure if that works, if not then you should implement it.

PR #526, opened 2018-10-18
Update Patches/Core/Stats/Stats.xml

XeoNovaDan:


Patches/Core/Stats/Stats.xml
```diff
@@ -56,28 +56,40 @@
...
 
 	<Operation Class="PatchOperationReplace">
-		<xpath>*/StatDef[defName="ShootingAccuracy"]/label</xpath>
+		<xpath>*/StatDef[defName="ShootingAccuracyPawn"]/label</xpath>
 		<value>
 			<label>weapon handling</label>
 		</value>
 	</Operation>
 
+	<Operation Class="PatchOperationSequence">
```
NoImageAvailable: Why are we running the same operation twice?

XeoNovaDan: First operation is PatchOperationTest and second is PatchOperationRemove

NoImageAvailable: Ah didn't spot that. Only issue then is this PR should go into the development branch.

XeoNovaDan: Alright, corrected

PR #535, opened 2018-11-09
Use CE mass to override vanilla mass, use CE cooldown to override vanilla cooldown

jinlan:


NoImageAvailable:
There is no such thing as CE mass/cooldown, what is the point of this?

Alicecomma:
I think it's related to the error logs, where a lot of errors mention that "Mass" or "Cooldown" have been defined twice.

jinlan:
Both CE and vanilla "Weapon mass" and "ranged Weapon Cooldown" state. They conflict

NoImageAvailable:
It did in previous alphas though, what changed? I'd say the way to fix this should be to restore previous functionality, i.e. override node if present in patched def, add it in otherwise. 

This PR would clear out mass/cooldown in cases where the CE patches doesn't touch those values, which is not what we want.

PR #537, opened 2018-11-09
More XML Error Fixes

Alicecomma:
- moved duplicate graphicData to match 1.0 definitions
- made ai_chillDestination false so pawns don't wander on barbed wire


NoImageAvailable:
> made ai_chillDestination false so pawns don't wander on barbed wire

You mean you don't go stargazing and cushion your head on some nice, soft razorwire?

PR #540, opened 2018-11-11
Remove Sniper & Heavy turret

Alicecomma:
- no research
- removed from buildings_security


Alicecomma:
Similar to #530 but without constructionSkillPrerequisite, no changes to Turret_MiniTurret cost and energy consumption, and no splitting of the weapon XML's into 1.0-like structure

PR #575, opened 2018-11-16
Remove Armor from Crashlanded

N7Huntsman:
Removed armor from Crashlanded


N7Huntsman:
Understood. Taking a closer look at it, seems the safest way to remove the items is with a PatchOpRemove for each item, then? I don't think I can specify values for Remove, can I?

PR #587, opened 2018-11-20
Some fixes

kpqi5858:
Contains various fixes


Source/CombatExtended/CombatExtended/AmmoInjector.cs
```diff
@@ -61,8 +61,8 @@ public static bool InjectAmmos()
 	            foreach (ThingDef def in DefDatabase<ThingDef>.AllDefsListForReading)
 	            {
 	            	if (def.IsWeapon
-	            	    && (def.generateAllowChance <= 0
-	            	        || def.tradeability == Tradeability.Buyable
+	            	    && (def.generateAllowChance > 0
+	            	        || def.tradeability.TraderCanSell()
```
NoImageAvailable: This seems like it would break determination of availability, what's the reasoning behind this change?

Source/CombatExtended/CombatExtended/Loadouts/Dialog_ManageLoadouts.cs
```diff
@@ -72,7 +72,7 @@ public Dialog_ManageLoadouts(Loadout loadout)
             	CurrentLoadout = loadout;
             SetSource(SourceSelection.Ranged);
             doCloseX = true;
-			doCloseButton = true;
+			//doCloseButton = true; //Close button is awkward
```
NoImageAvailable: Reasoning behind this?

Source/CombatExtended/CombatExtended/AmmoInjector.cs
kpqi5858: That method means tradeability == Buyable || tradeability == All

Source/CombatExtended/CombatExtended/Loadouts/Dialog_ManageLoadouts.cs
kpqi5858: Close button draws over item selection row, and awkward

Source/CombatExtended/CombatExtended/AmmoInjector.cs
kpqi5858: And why generateAllowChance should less than or equal 0 to be a weapon?

Alicecomma: Good change.

These changes by kpqi5858 are as they should be. The code present right now is actually wrong, because it only picks weapons which aren't generatable.

PR #665, opened 2019-02-06
Boomalope Crash Fix - Update JobDriver_Hunt.cs 

General1337:
Fixed Boomalope crash by having Pawns shoot downed dangerous animals instead of causing them to get caught in an infinite loop and crash.

### DIFF
--- a/Source/CombatExtended/CombatExtended/Jobs/JobDriver_Hunt.cs
+++ b/Source/CombatExtended/CombatExtended/Jobs/JobDriver_Hunt.cs
@@ -68,10 +68,18 @@ namespace CombatExtended
             yield return Toils_Jump.JumpIfTargetDespawnedOrNull(VictimInd, startCollectCorpse);
 
             var startExecuteDowned = Toils_Goto.GotoThing(VictimInd, PathEndMode.Touch).JumpIfDespawnedOrNull(VictimInd, startCollectCorpse);
+
+            var startExecuteDownedRanged = Toils_Combat.CastVerb(VictimInd);
+
+            yield return Toils_Jump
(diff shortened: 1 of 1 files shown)


LockdownX7:
Tried this fix in my current colony for about an hour, with lots of hunting jobs going around. Seems to fix the issue without breaking anything else.

Source/CombatExtended/CombatExtended/Jobs/JobDriver_Hunt.cs
```diff
@@ -68,10 +68,18 @@ protected override IEnumerable<Toil> MakeNewToils()
             yield return Toils_Jump.JumpIfTargetDespawnedOrNull(VictimInd, startCollectCorpse);
 
             var startExecuteDowned = Toils_Goto.GotoThing(VictimInd, PathEndMode.Touch).JumpIfDespawnedOrNull(VictimInd, startCollectCorpse);
+
+            var startExecuteDownedRanged = Toils_Combat.CastVerb(VictimInd);
```
NoImageAvailable: Isn't this redundant? If the jump to startExecuteDowned fails to trigger it will proceed to shoot it anyways.

General1337: Each toil is always called, so without startExecuteDownedRanged the pawn will always try to approach a downed target. This forces the pawn to fire at the downed target if the toil condition is met.

NoImageAvailable: >if the toil condition is met

Which doesn't happen, hence my point.

NoImageAvailable:
Equivalent included in https://github.com/NoImageAvailable/CombatExtended/commit/2bf72dbac3eedff1a465f77e1c23c6a53fa0128f

PR #669, opened 2019-02-16
few changes in gear tab and loadouts menu

Psionix:
loadouts menu
-added pause when menu is opened.
-fixed low fps.
-some minor visual changes.
-plain(in-code) strings replaced by translatable strings.
gear tab
-added text overlay on weight and bulk bars.
-added detailed tooltip when hovering cursor over armor values оf creature.
-added item weight values.


Psionix:
Also fixing displaying bullets in loadouts menu.

PR #680, opened 2019-02-28
fix #679

### LINKED ISSUE #679: Bulk/mass bars gone rogue
They say a picture is worth a thousand words, so: https://i.imgur.com/ttkb5iM.png

Happened when the pawn underwent anesthesia for an operation, and their mass capacity dropped to 0. Gear tooltip shows quite the move speed penalty: https://i.imgur.com/H7hy5KQ.png

Also on a vaguely related note: in the Assign window, the hover-tooltips for the Mass and Bulk bars are switched around. EDIT: Not just the tooltips, it's the bars themselves that need to switch places to match the window's header.

Psionix:


Source/CombatExtended/CombatExtended/MassBulkUtility.cs
```diff
@@ -24,9 +24,11 @@ public static float BaseCarryBulk(Pawn p)
         }
 
         public static float MoveSpeedFactor(float weight, float weightCapacity)
-        {
-            return Mathf.Lerp(1f, 0.75f, weight / weightCapacity);
-        }
+        {
+			float t = weight / weightCapacity;
+			if (float.IsNaN(t)) t = 0f;
```
NoImageAvailable: Should use min value instead of 0 to avoid AI derpiness.

Psionix: done

PR #683, opened 2019-03-02
add stuff and quality to loadbearing gear

### LINKED ISSUE #279: Backpacks should have material/quality properties
Right now backpacks are always made from cloth and have no quality property.  This means they cant be stored in stockpiles with quality restrictions or made from leather/wool/devilstrand. Treating backpacks like other clothing with quality and material properties would fix these.

ymber:
fixes #279 

Testing in game didn't show any problems with high armor stuff creating armored backpacks.


NoImageAvailable:
There's a reason #279 was closed. Since quality and stuff don't actually affect the functionality of the item, only it's market value there's a slew of potentially broken interactions like raiders blowing their entire budgets on legendary hyperweave backpacks and coming in naked that would have to be tested against for a really minor gain.

At most I'd PR a quality only version of this because that at least has relatively minor effects on market value without Stuff compounding it and is causing legitimate issues with the outfit/stockpile systems.

ymber:
Didn't catch that the first time. Testing again there's a noticeable drop in gear quality when they bring expensive loadbearing gear. I'll move the stuff part of this out to a mod and see if I can do something with raid point allocation later.

PR #697, opened 2019-04-04
Added FSX Distillation

N7Huntsman:
Add recipe to Biofuel Refinery to allow conversion of Chemfuel to FSX at ratio of 5:1. While not as economical as raising boomalopes/boomrats, it makes FSX available to colonies in biomes where they do not naturally occur or cannot be ranched (such as extremely cold climates).


NoImageAvailable:
As explained elsewhere not gonna make this part of the main mod. Should be an optional addon instead.

PR #701, opened 2019-04-15
added support to Extended Storage Extended

jinlan:


NoImageAvailable:
There already is an Extended Storage patch

PR #702, opened 2019-04-19
Fix for continous NPE

KiameV:
Was constantly getting NPE when JobGiver_TakeAndEquip.GetPriorityWork was called. This will fix the problem.

I also noticed GetPriorityWork was being called many times in TryGiveJob. Changed the code to only call it once.


Source/CombatExtended/CombatExtended/Jobs/JobGiver_TakeAndEquip.cs
```diff
@@ -81,7 +81,8 @@ private WorkPriority GetPriorityWork(Pawn pawn)
                 
             	if (hasWeaponTags
                   && primaryAmmoUser.parent.def.weaponTags.Any(pawn.kindDef.weaponTags.Contains)
-            	  && loadoutPropertiesExtension?.primaryMagazineCount != FloatRange.Zero)
+                  && loadoutPropertiesExtension != null
+                  && loadoutPropertiesExtension?.primaryMagazineCount != FloatRange.Zero)
```
NoImageAvailable: Why check for null twice?

KiameV:
loadoutPropertiesExtention? apparently equals FloatRange.Zero so the if statement succeeds when primaryMagazineCount is not set outright

Source/CombatExtended/CombatExtended/Jobs/JobGiver_TakeAndEquip.cs
KiameV: I think you mean the ? on the line after the null check. Removed.

PR #737, opened 2019-07-31
Allow manned turrets to be reloaded via haul job

LockdownX7:
Mannable turrets can be reloaded through hauling job or right-click context menu, as long as the turret isn't in use.


NoImageAvailable:
How does this behave with in-progress haul jobs and state changes (e.g. turret starts being manned, deconstructed, etc.)?

LockdownX7:
Just like auto-turrets, mannable turrets become reserved when there's a job involving them. Other pawns can't interact with the turret while ammo is being hauled to it, and if a manual order is given, the haul job is cancelled.

PR #738, opened 2019-07-31
Fix #724

### LINKED ISSUE #724: Pawns try to fire with empty mag before reloading
Not sure if this is intended, but pawns often waste time aiming at targets while their mag is empty, only reloading after done aiming. This affects all bolt-action and semi-automatic weapons. Automatics are only affected if their burst count exactly matches the number of rounds left in the mag.

In particular, this severely cripples Inferno cannons due to their single-shot nature, as the centipede will fire one round, go through the process of aiming another shot, realize it's empty, reload a new one, then finally fire a second round for real.

LockdownX7:
Fixes the issue where pawns go through an aiming cycle while their magazine is already empty.

### DIFF
--- a/Source/CombatExtended/CombatExtended/Verbs/Verb_ShootCE.cs
+++ b/Source/CombatExtended/CombatExtended/Verbs/Verb_ShootCE.cs
@@ -94,41 +94,9 @@ namespace CombatExtended
         // Whether our shooter is currently under suppressive fire
         private bool IsSuppressed => ShooterPawn?.TryGetComp<CompSuppressable>()?.isSuppressed ?? false;
 
-        //Checks if magazine-fed weapon needs to reload before able to shoot again (out-of-ammo state is automatically handled by TryStartReload)
-        private bool NeedsMagazineReload => CompAmmo != null && CompAmmo.HasMagazine && CompAmmo.CurMagCount == 0;
-
-        //Mag-less weapons need separate handling for out-of-ammo state, as TryStartReload does nothing for them
-        private bool IsMaglessWeaponOutOfAmmo => CompAmmo != null && !CompAmmo.HasMagazine && CompAmmo.UseAmmo && !CompAmmo.HasAmmo;
-
-        private bool IsAttacking => ShooterPawn?.CurJobDef == JobDefOf.AttackStatic || ShooterPawn?.stances.curStance is Stance_Warmup;
-
         #endregion
 
         #region Methods
-
-        public override bool Available()
-        {
-            if (!base.Available())
-            {
-                return false;
-            }
-
-            if (IsAttacking)
-            {
-                if (NeedsMagazineReload)
-                {
-                    CompAmmo?.TryStartReload();
-                    return false;
-                }
-                if (IsMaglessWeaponOutOfAmmo)
-                {
-                    CompAmmo?.DoOutOfAmmoAction();
-                    return false;
-                }
-            }
-
-            return true;
-        }
 
         /// <summary>
         /// Handles activating aim mode at the start of the burst
@@ -254,10 +222,6 @@ namespace CombatExtended
                         }
                     }
                     return CompAmmo.Notify_PostShotFired();
-                }
-                if (NeedsMagazineReload)
-                {
-                    CompAmmo?.TryStartReload();
                 }
                 return true;
             }
--- a/Source/CombatExtended/CombatExtended/Verbs/Verb_ShootCE.cs
+++ b/Source/CombatExtended/CombatExtended/Verbs/Verb_ShootCE.cs
@@ -94,9 +94,41 @@ namespace CombatExtended
         // Whether our shooter is currently under suppressive fire
         private bool IsSuppressed => ShooterPawn?.TryGetComp<CompSuppressable>()?.isSuppressed ?? false;
 
+        //Checks if magazine-fed weapon needs to reload before able to shoot again (out-of-ammo state is automatically handled by TryStartReload)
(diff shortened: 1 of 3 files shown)


Source/CombatExtended/CombatExtended/Verbs/Verb_ShootCE.cs
```diff
@@ -94,10 +94,42 @@ protected override float SwayAmplitude
         // Whether our shooter is currently under suppressive fire
         private bool IsSuppressed => ShooterPawn?.TryGetComp<CompSuppressable>()?.isSuppressed ?? false;
 
+        //Checks if magazine-fed weapon needs to reload before able to shoot again (out-of-ammo state is automatically handled by TryStartReload)
```
NoImageAvailable: Why not use the properties already included in CompAmmo?

Source/CombatExtended/CombatExtended/Verbs/Verb_ShootCE.cs
```diff
@@ -94,10 +94,42 @@ protected override float SwayAmplitude
...
+        //Mag-less weapons need separate handling for out-of-ammo state, as TryStartReload does nothing for them
+        private bool IsMaglessWeaponOutOfAmmo => CompAmmo != null && !CompAmmo.HasMagazine && CompAmmo.UseAmmo && !CompAmmo.HasAmmo;
+
+        private bool IsAttacking => ShooterPawn?.CurJobDef == JobDefOf.AttackStatic || ShooterPawn?.stances.curStance is Stance_Warmup;
+
         #endregion
 
         #region Methods
 
+        public override bool Available()
```
NoImageAvailable: This should be handled via the base class

Source/CombatExtended/CombatExtended/Verbs/Verb_ShootCE.cs
```diff
@@ -94,10 +94,42 @@ protected override float SwayAmplitude
...
+            if (IsAttacking)
+            {
+                if (NeedsMagazineReload)
+                {
+                    CompAmmo?.TryStartReload();
+                    return false;
+                }
+                if (IsMaglessWeaponOutOfAmmo)
+                {
+                    CompAmmo?.DoOutOfAmmoAction();
```
NoImageAvailable: OutOfAmmoAction is already part of TryStartReload

LockdownX7: It is, but if you call TryStartReload on something like a bow, the function returns instantly because there isn't a magazine to reload, and the pawn will just sit there staring at their target, instead of switching to their next weapon like they should.

Source/CombatExtended/CombatExtended/Verbs/Verb_ShootCE.cs
LockdownX7: If you mean why not use the CanBeFiredNow property: because that one always returns true when the ammo system is disabled, even though weapons still need to reload when their magazine goes empty. Having checked again, I still can't find a property that does exactly what I need for NeedsMagazineReload.

Source/CombatExtended/CombatExtended/Verbs/Verb_ShootCE.cs
LockdownX7: I put this in Verb_ShootCE because there's also an added reload check at the end of TryCastShot(), and it made sense to me that all the reload logic should be contained in the same class. With that in mind, should the TryCastShot() reload check also be moved into LaunchProjectile?

Source/CombatExtended/CombatExtended/Verbs/Verb_ShootCE.cs
NoImageAvailable: That's a bug in TryStartReload then and should be fixed there, not here.

Source/CombatExtended/CombatExtended/Verbs/Verb_ShootCE.cs
NoImageAvailable: The base class already has a reload check though.

Source/CombatExtended/CombatExtended/Verbs/Verb_ShootCE.cs
NoImageAvailable: There's a slew of properties for all combinations of magazine and ammo states that should make all these additions redundant.

LockdownX7: After the change to make TryStartReload() handle out-of-ammo state properly for bows/etc, I was able to simplify the code a fair bit, but I'm still having to use a large combination of AmmoUser properties to know whether TryStartReload() should be called. Here's my dilemma:

For magazine weapons, we just need to know if the mag count is at 0. It doesn't matter if there's ammo in the inventory, or if the ammo system is disabled entirely, the resulting action is always the same: call TryStartReload() so it either puts a fresh mag in, or unequips the weapon. In code terms, that's (HasMagazine && CurMagCount == 0).

For no-mag weapons, it's slightly different: We do care about ammo in the inventory, because the weapon feeds directly off it. But only if the ammo system is on. That translates to (!HasMagazine && UseAmmo && !HasAmmo).

In summary, a property is needed that resolves ((HasMagazine && CurMagCount == 0) || (!HasMagazine && UseAmmo && !HasAmmo)). Is there one in CompAmmoUser that does such a thing? No...  !CanBeFiredNow comes really close, but it assumes that magazines never need to reload when the ammo system is disabled, so it doesn't make the cut. (That might well be a bug for all I know, though)

So where do I go from here? Is the logic of CanBeFiredNow intentional, or do I need to fix it? If it's intended, should I make a separate new property in CompAmmoUser for the needed purpose?

PR #749, opened 2019-08-22
Obscure colony start fix #1

### LINKED ISSUE #1: Remove pants-pissing
Need to remove urination mechanic from highest level suppression. Potentially replace pants-pissing tale with something more tasteful. Might be viable to replace with pawn entering catatonic state (mental break?).

Alicecomma:
Fix for ancient danger "Halfeaten" pawns' armor being destroyed
(happens like once in fifty game starts)
TakeDamage()->Apply->ApplyToPawn->ApplyDamageToPart->ArmorReroute->GetAfterArmorDamage

### DIFF
--- a/Source/CombatExtended/CombatExtended/ArmorUtilityCE.cs
+++ b/Source/CombatExtended/CombatExtended/ArmorUtilityCE.cs
@@ -124,7 +124,7 @@ namespace CombatExtended
                         // Hit was deflected, convert damage type
                         //armorReduced = true;
                         dinfo = GetDeflectDamageInfo(dinfo, hitPart);
-                        if (app == apparel.ElementAtOrDefault(i))   //Check whether the "deflecting" apparel is still in the WornApparel - if not, it has been destroyed by the damage
+                        if (app == apparel.ElementAtOrDefault(i))   //Check whether the "deflecting" apparel is still in the WornApparel - if not, the next loop checks again and errors out because the index is out of range
                            i++;    // We apply this piece of apparel twice on conversion, this means we can't use deflection on Blunt or else we get an infinite loop of eternal deflection
                     }
                     if (dmgAmount <= 0)
--- a/Source/CombatExtended/CombatExtended/ArmorUtilityCE.cs
+++ b/Source/CombatExtended/CombatExtended/ArmorUtilityCE.cs
@@ -115,13 +115,17 @@ namespace CombatExtended
                 // Apparel is arranged in draw order, we run through reverse to go from Shell -> OnSkin
                 for (int i = apparel.Count - 1; i >= 0; i--)
                 {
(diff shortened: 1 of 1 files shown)


Alicecomma:
Easiest way to test for the fix in bulk is to set Mods\Core\Defs\MapGeneration\BasePlayerMapGenerator.xml's ScatterShrines to countPer10kCellsRange min: 10, max: 15.

Source/CombatExtended/CombatExtended/ArmorUtilityCE.cs
```diff
@@ -115,13 +115,17 @@ public static DamageInfo GetAfterArmorDamage(DamageInfo originalDinfo, Pawn pawn
...
+
+                    if (app != null
+                        && app.def.apparel.CoversBodyPart(hitPart)
+                        && !TryPenetrateArmor(dinfo.Def, app.GetStatValue(dinfo.Def.armorCategory.armorRatingStat), ref penAmount, ref dmgAmount, app))
                     {
                         // Hit was deflected, convert damage type
                         //armorReduced = true;
                         dinfo = GetDeflectDamageInfo(dinfo, hitPart);
-                        i++;    // We apply this piece of apparel twice on conversion, this means we can't use deflection on Blunt or else we get an infinite loop of eternal deflection
+                        if (app == apparel.ElementAtOrDefault(i))   //Check whether the "deflecting" apparel is still in the WornApparel - if not, it has been destroyed by the damage
```
NoImageAvailable: If the goal is to check for destruction, why not use .Destroyed?

Alicecomma: What this method does is check that the reference to the apparel (app) is still present at the position apparel[i]. The reference will remain valid, but the list apparel will not contain app anymore.

I have no idea why a piece of apparel is iterated twice in this method, but it is the source of the error as it will check the apparel[i] twice, and if i is equal to the length of WornApparel (e.g outer layer of the pawn is destroyed), the method crashes. This only happens when the type of damage done to the apparel isn't sharp damage (I think?) and a whole bunch of other factors - which comes down to the error only popping up whenever a backpack (which is always the outer layer of the pawn) is destroyed while the method determines a deflection has occurred, then the apparel is checked twice and the method figures it ain't there anymore.

E.g I don't know. Maybe it is destroyed, maybe not. It's just not in the list anymore.

Alicecomma: Latest commit changes the comment to not mention "destroyed" because I have no idea but it works

NoImageAvailable: The repeat iteration happens on deflection: sharp damage is converted to blunt and reapplied to the same piece of armor in order to check against it's blunt rating.

PR #750, opened 2019-08-22
Obscure colony start fix #2

### LINKED ISSUE #2: Fix ammo armor penetration
Armor penetration algorithm gets projectile from firing weapon's def, meaning armor calculation doesn't take into account actual projectile fired. Need to change dinfo back to passing projectile to armor calculation.

Alicecomma:
Fix for ancient danger "Halfeaten" pawns spawning without arms, then
their loadout adding a shield and a warning telling us they can't wear
the shield because they don't have arms


Alicecomma:
Easiest way to test for the fix in bulk is to set Mods\Core\Defs\MapGeneration\BasePlayerMapGenerator.xml's ScatterShrines to countPer10kCellsRange min: 10, max: 15.

PR #755, opened 2019-08-25
Additional stat displays in ranged/melee recipes and ammoThings

Alicecomma:
- Display Caliber in the information box in Bills (req.Def as ThingDef
instead of req.Thing.Def)

- Display magazine in Bill info box (req.Def as ThingDef instead of
req.Thing.Def)

- Display CORRECT melee armor penetration in Bill info box
(req.Thing.(..).MeleePenetrationFactor ?? 1f to default to 1f)

- Display AmmoSetDefs (Guns) which an AmmoThing (or AmmoDef) can be used
in, using StatDef Ammoset.

### DIFF
--- a/Defs/Stats/Stats_Basics_Inventory.xml
+++ b/Defs/Stats/Stats_Basics_Inventory.xml
@@ -11,5 +11,15 @@
     <toStringStyle>FloatTwo</toStringStyle>
     <displayPriorityInCategory>10</displayPriorityInCategory>
   </StatDef>
+  
+  <StatDef>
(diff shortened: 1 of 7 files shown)


Alicecomma:
The way AmmoUser currently works, this method is sufficient. However, if in the future it's possible to modify the magazineSize, ammoSetDef or tools of a Thing (without modifying the ThingDef, through the CompAmmoUser), these methods should consider that factor.

Alicecomma:
Additional commits coming in--

PR #757, opened 2019-08-25
Remove prometheum hediff from napalm

LockdownX7:
Uses the standard ~Burn~ Flame def instead, as napalm is crafted with good ol' chemfuel.


Alicecomma:
To be fair the prometheum hediff should be on the napalm because it just makes it so the fuel sticks to the target, making it more flammable

NoImageAvailable:
It also causes the target to self-ignite, meaning it can't be put out without firefoam. The concept behind prometheum is that it's pyrophoric, napalm is not.

PR #759, opened 2019-08-25
Fixed tradeable rounds, destroyOnDrop=false for some ammo

Alicecomma:
- Mechanoid ammo had tradeability = Sellable, which gave warnings. Someone removed the warnings by setting destroyOnDrop=false, but this meant the ammo still existed in the debug menu when the ammo system was off (and could still be found in say mission rewards).

- A new part was written in the AmmoInjector and new tradeTags were allowed so tradeability = Sellable is set when ammo is on, but it's set to None when ammo is off, fixing the issue.

- 90mm Cannon had an incorrect crafting base (GrenadeRecipeBase) which meant the crafting recipe still displayed in the machining table even with ammo turned off.

### DIFF
--- a/Defs/Ammo/Advanced/12x64mmCharged.xml
+++ b/Defs/Ammo/Advanced/12x64mmCharged.xml
@@ -30,8 +30,9 @@
       <li>Ammo12x64mmCharged</li>
     </thingCategories>
     <stackLimit>200</stackLimit>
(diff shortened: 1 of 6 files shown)


Alicecomma:
This fix makes it so some of the rounds which earlier on showed with the ammo system disabled are now correctly not there anymore

Issue #761, opened 2019-08-25
Mortar-mounted pawns with a Shieldbelt can't aim at the ground and keep the gun aiming without shooting


Alicecomma:
- Prevent pawns with a shieldbelt to mount mortars;

OR

- Allow pawns with a shieldbelt to shoot when mounted.

Alicecomma:
Original bug report on Steam (https://steamcommunity.com/profiles/76561198238877256):

CynSaje 14 Aug @ 2:02am 
Hi, I don't know if it is intended, but I discovered that if you have a shield belt while manning a mortar, the mortars go in a infinite reloading loop.

I don't know if this problem is present in vanilla though, but I never had it before installing this mod.

NoImageAvailable:
Shield belts preventing turret operation is a vanilla feature.

Alicecomma:
This isn't true:
![image](https://user-images.githubusercontent.com/4662406/63693512-8fdc6e80-c814-11e9-9233-eadd69a6b5f2.png)

Alicecomma:
Same when drafted with the shield visibly active:
![image](https://user-images.githubusercontent.com/4662406/63693778-37f23780-c815-11e9-9241-13b61850d883.png)

Alicecomma:
So the problem with the Harmony postfix currently is that it's logically impossible to form the expression (!A && !C) || B which is needed to validate the operation being allowed:

```

            /*
             *  A   B   C   D       E
             *  0   0   0   1       1
             *  0   0   1   1       0
             *  0   1   0   1       1
             *  0   1   1   1       1
             *  1   0   0   0       0
             *  1   0   1   0       0
             *  1   1   0   1       1
             *  1   1   1   1       1
             * 
             * A = verb is Verb_LaunchProjectile
             * B = ReachabilityImmediate(..)
             * C = verb is Verb_LaunchProjectile_CE
             * D = (!A) || B
             * E = (!A && !C) || B
             * 
             * We know A, C and D. Can we get E?
             * 
             *  A   C   D       F
             *  0   0   1       1
             *  0   1   1       0
             *  0   0   1       1
             *  0   1   1       1   ==> Duplicate of line 2, but a different result
             *  1   0   0       0
             *  1   1   0       0
             *  1   0   1       1
             *  1   1   1       1
             *  
             *  So, we can't get E from A, C and D alone. We need B.
             */
```

[...]

PR #767, opened 2019-08-27
Disable shield belt while operating turrets

LockdownX7:
As discussed in #761, this allows shield-belt wearers to operate manned turrets (matching vanilla behaviour), while also disabling the shield effect for the duration, for balance reasons.


LockdownX7:
Redone the PR, the initial solution wasn't the best approach.

Source/CombatExtended/CombatExtended/Verbs/Verb_LaunchProjectileCE.cs
```diff
@@ -497,15 +497,16 @@ public virtual bool CanHitTargetFrom(IntVec3 root, LocalTargetInfo targ, out str
...
                 {
                     report = "IsIncapableOfViolenceLower".Translate(ShooterPawn.Name.ToStringShort);
                     return false;
                 }
 
                 // Check for apparel
-                if (ShooterPawn.apparel != null)
+                //pawns can use turrets while wearing shield belts, but the shield is disabled for the duration via Harmony patch (see Harmony-ShieldBelt.cs)
+                bool isTurretOperator = caster.def.building?.IsTurret ?? false;
+                if (ShooterPawn.apparel != null && !isTurretOperator)
```
NoImageAvailable: This is gonna enable turret operation regardless of any apparel, not just shield belts.

LockdownX7: My thinking was that mod-added apparel would implement their own Harmony postfixes in their CE patch, but now that you mention it, I can't really make that assumption: some apparel might simply be intended to not allow firing at all, rather than having something that can be toggled off to allow it.

LockdownX7:
Wait, the logic on that apparel is wrong, will fix soon.

Issue #112, opened 2017-03-28
Unload weapons on the ground.


ProfoundDarkness:
1. In issue #100 got the idea of allowing guns not in a pawn's inventory to be unloaded.  The core idea being to allow a **right click menu for guns on the ground with ammo to be unloaded**.
----
2. ~Stepping further out and probably reliant on the code for the above, would be **automatic unload when smelting/destroying a gun**~ (#768)
----
3. potentially a way to automate weapon unloading kind of like how one can mark a corpse for stripping, marking a building for destruction, or marking rock wall for mining...

NoImageAvailable:
Don't designator-adding mods have hard incompatibilities with each other? While the idea is sound I'd rather not require a special patch for stuff like vein miner for something minor like this.

Alicecomma:
RE: Smelting guns (#764):

https://steamcommunity.com/workshop/filedetails/discussion/1631756268/1639794651167212372/

Could be done by a Harmony patch of the IEnumerable Thing.SmeltProducts. Not sure how, I'm not great with Harmony patches.

Alicecomma:
![image](https://user-images.githubusercontent.com/4662406/63898611-04243700-c9fa-11e9-97af-93b95683649b.png)

PR #768, opened 2019-08-28
PostFix SmeltProducts to add any loaded ammo #112

Alicecomma:
Partial fix for #112 - the smelting weapons part


Alicecomma:
The postfix should probably be in a new file "Harmony-Thing.cs" instead of HarmonyThingDef.cs.

Alicecomma:
Moved to new file Harmony-Thing.cs

PR #772, opened 2019-09-01
Tweak to GouRIMet Patch

N7Huntsman:
Fixed oversight in initial patch.


Alicecomma:
1000 commit party

image

PR #785, opened 2019-09-13
Fix cover leaning, skip collision checks at point blank

LockdownX7:


Source/CombatExtended/CombatExtended/Verbs/Verb_LaunchProjectileCE.cs
```diff
@@ -572,6 +572,7 @@ protected override bool TryCastShot()
                     targDist *= 2;  // Double to account for divide by 4 in ProjectileCE minimum collision distance calculations
                 projectile.minCollisionSqr = Mathf.Pow(targDist, 2);
                 projectile.intendedTarget = currentTarget.Thing;
+                projectile.targetCell = currentTarget.Cell;
```
NoImageAvailable: Is this actually producing appropriate results with pawns as opposed to using intendedTarget

LockdownX7: Haven't found any issues with it, either against single or multi-tile pawns. Nevertheless, I edited ProjectileCE so it uses intendedTarget if it's not null, only relying on targetCell if firing manually at the ground.

PR #787, opened 2019-09-15
Flammability info fixes

LockdownX7:
- Fixed error in animal info window due to null apparel
- Apparel flammability was showing as int.MinValue for pawns with zero coverage, added check on division.

### DIFF
--- a/Source/CombatExtended/CombatExtended/StatWorkers/StatWorker_Flammability.cs
+++ b/Source/CombatExtended/CombatExtended/StatWorkers/StatWorker_Flammability.cs
@@ -63,7 +63,7 @@ namespace CombatExtended
             var stringBuilder = new StringBuilder();
             stringBuilder.Append(explanation);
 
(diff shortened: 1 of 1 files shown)


Source/CombatExtended/CombatExtended/StatWorkers/StatWorker_Flammability.cs
```diff
@@ -74,7 +74,7 @@ public override string GetExplanationUnfinalized(StatRequest req, ToStringNumber
             if ((req.Thing is Pawn pawn))
             {
                 GetApparelAdjustFor(pawn, out var apparelFlammability, out var apparelCoverage);
-                apparelFlammability /= apparelCoverage;
+                apparelFlammability /= apparelCoverage != 0f ? apparelCoverage : 1f;
```
NoImageAvailable: Should just skip the entire calculation for pawns with null apparel

PR #791, opened 2019-09-22
Added patch for Gas Traps & Shell

N7Huntsman:
The patch includes recipes added conditionally, and the ammo category def file within the patch folder (let me know if this must change).

### DIFF
--- a/Patches/Gas Traps and Shells/Defs/AmmoCategories_Gas.xml
+++ b/Patches/Gas Traps and Shells/Defs/AmmoCategories_Gas.xml
@@ -1,0 +1,46 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<Defs>
+
+   <CombatExtended.AmmoCategoryDef>
+    <defName>GasToxic</defName>
+    <label>toxic gas</label>
+    <labelShort>Toxic</labelShort>
+    <description>This shell contains a potent, nerve agent which is extremely harmful to anyone exposed.</description>
+  </CombatExtended.AmmoCategoryDef>
+
+   <CombatExtended.AmmoCategoryDef>
+    <defName>GasRage</defName>
+    <label>rage gas</label>
+    <labelShort>Rage</labelShort>
+    <description>This shell contains a potent, specialized airborne agent that induces blind, uncontrollable rage.</description>
+  </CombatExtended.AmmoCategoryDef>
+
+   <CombatExtended.AmmoCategoryDef>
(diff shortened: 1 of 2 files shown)


Patches/Gas Traps and Shells/Defs/AmmoCategories_Gas.xml
```diff
@@ -0,0 +1,46 @@
...
+    <label>rage gas</label>
+    <labelShort>Rage</labelShort>
+    <description>This shell contains a potent, specialized airborne agent that induces blind, uncontrollable rage.</description>
+  </CombatExtended.AmmoCategoryDef>
+
+   <CombatExtended.AmmoCategoryDef>
+    <defName>GasTear</defName>
+    <label>tear gas</label>
+    <labelShort>CN</labelShort>
+    <description>This shell contains a potent, specialized airborne agent that attacks the sensitive membranes of the eyes and nose, making it difficult to sea or breathe.</description>
```
NoImageAvailable: I sea a number of spelling mistakes here. Also, why add the ammo dynamically but not the categories?

N7Huntsman: I'm not seeing the spelling mistakes, unless you're referring to the somewhat atypical grammar of, "...to anyone exposed." I suppose I can change that if you like.

Regarding the categories, in so far as I was aware they don't affect anything if they're loaded and not used, so I figured it just saved the load time for another patch. However, I didn't realize Rimworld actually threw an error for a Def file in the Patches folder (I figured they were mostly just for organization on the user side of things and that RW only cared about the file itself), so I'll just close this PR and go back and make it a conditional patch for the sake of keeping the whole thing nice and neat within a single folder under Patches.
 
For the recipes, my reasoning was that since they're part of the 81mm ammo set, the recipes will still appear regardless of whether or not the Gas Shells & Traps mod is loaded, which is obviously problematic.

N7Huntsman:
Will revise grammar and Ammo Categories into conditional patch.

Patches/Gas Traps and Shells/Defs/AmmoCategories_Gas.xml
NoImageAvailable: One patch that immediately cancels out when it detects a mod not loaded takes less time to load than several additional defs. That's why ammo-related defs that are never going to be used outside that particular mod should be added in dynamically.

PR #805, opened 2019-09-28
Unarmed damage changes

LockdownX7:
Humans scale from 0% to 300% damage depending on their melee skill level, which, with base fist damage being 3, puts them in the 0-9 damage range. Animals don't have melee skill, so they always do 100% of base.

Also added latest melee stats for megaspiders and scythers, and validation checks in melee damage statworker.


LockdownX7:
Changes to dev branch and spreadsheets have made this PR obsolete, closing.

PR #822, opened 2019-10-07
Fix Apparello patch

LockdownX7:


LockdownX7:
Jeez, GitHub decided to treat it as a monolith block of changes. What I've changed is wrap the li's into an operations element, and change the third-to-last operation to Add, as those apparels don't have StuffEffectMultiplierArmor defined.

LockdownX7:
Already fixed in dev, closing.

PR #833, opened 2019-10-16
Fix damage display in info windows

LockdownX7:
image
image


NoImageAvailable:
Potentially problematic with tools that have more than one capacity but good enough for now

PR #841, opened 2019-10-26
Update AssemblyInfo version

JonnyNova:


NoImageAvailable:
Isn't this gonna break any third party dependencies?

JonnyNova:
To the best of my knowledge, it depends if they are explicitly checking the version of the assembly after loading it with [Assembly.Load](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.load?view=netframework-3.5)

Though, the current version information is wrong and won't give any useful clues to anyone interested in checking what the version actually is.

I'm checking for version information when doing manual integrations here: https://github.com/rimworld-frontierdevelopment/Core/blob/master/Source/General/ModIntegration.cs

PR #844, opened 2019-10-30
New armor system displays for gear tab

LockdownX7:
image


LockdownX7:
I suppose it should. Fixed.
image

PR #848, opened 2019-11-02
Arachne Race CE 1.6 patch

Magichp:
Base mod:   https://steamcommunity.com/sharedfiles/filedetails/?id=1589454687

CE Patch:    https://steamcommunity.com/sharedfiles/filedetails/?id=1904384309

Balance is probably a bit rough, but this was intended as an OP race (regenerating organs and lost limbs for example) so I guess it fits ^^


NoImageAvailable:
This is just a PR for the CE master branch, not any patch

PR #874, opened 2019-11-09
Kurin CE patch

Tostov:
Should complete patch for _Kurin, The Three Tailed Fox_ .

I didn't touch the noose trap, but it doesn't seem to break anything anyway.

### DIFF
--- a/Patches/Kurin/ThingDefs_Races/Race_DRNTF.xml
+++ b/Patches/Kurin/ThingDefs_Races/Race_DRNTF.xml
@@ -1,0 +1,124 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<Patch>
+  <Operation Class="PatchOperationSequence">
+	<operations>
+	  <li Class="PatchOperationFindMod">
+			
+		<mods><li>Kurin, The Three Tailed Fox</li></mods>
+			
+		<match Class="PatchOperationSequence">
+		<operations>
+			
+			<li Class="PatchOperationConditional">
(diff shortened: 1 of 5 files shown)


Patches/Kurin/ThingDefs_Races/Race_DRNTF.xml
```diff
@@ -0,0 +1,124 @@
...
+			 <xpath>/Defs/AlienRace.ThingDef_AlienRace[defName="DRNTF_Race"]/alienRace/raceRestriction/whiteApparelList</xpath>
+				  <value>
+					<li>Apparel_TacVest</li>
+					<li>Apparel_Backpack</li>
+					<li>Apparel_BallisticShield</li>
+					<li>Apparel_MeleeShield</li>
+					<li>Apparel_GasMask</li>
+					<li>Apparel_ImprovGasMask</li>
+				  </value>
+			</li>
```
LockdownX7: You're missing the tribal backpack. (assuming there isn't a specific reason Kurins can't wear those)

LockdownX7 (changes requested):

Patches/Kurin/ThingDefs_Races/Race_DRNTF.xml
Tostov: Ahh dang, I always forget about it. I'll add it in to this one. Dragonian patch will be missing it too, so I'll put it in there as well.

LockdownX7 (approved):

PR #875, opened 2019-11-09
Dragonian CE patch

Tostov:
Should complete patch for _Dragonian Race_


LockdownX7 (changes requested):
Seeing as this patch is for `Dragonian Race`, the base folder name should reflect that, so it's not confused with the similarly named `Dragonian` by the same author.

Tostov:
Folder renamed/files moved.

LockdownX7 (approved):

PR #885, opened 2019-11-12
Girls frontline Weapon pack patch + ammodef

Tostov:
Patch and ammo def for Girls' frontline Weapon Pack.
Included 15.2x169mm ammo def since it is techincally a real caliber.
Ammo spec:
Overall Length: 207mm
Base case diameter: 26mm
Bullet Case Mass: 150g
Bullet Diameter (sabot): 5.5mm
Muzzle Velocity: 1450m/s
Bullet Mass (sabot):19.5g
Penetration is guesswork based on 40mm RHA at 1000m, which seems to be the most commonly cited penetration of the gun.

### DIFF
--- a/Defs/Ammo/HighCaliber/15.2x169mm.xml
+++ b/Defs/Ammo/HighCaliber/15.2x169mm.xml
@@ -1,0 +1,124 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<Defs>
+
+  <ThingCategoryDef>
+    <defName>Ammo152x169mm</defName>
+    <label>15.2x169mm</label>
+    <parent>AmmoHighCaliber</parent>
+    <iconPath>UI/Icons/ThingCategories/CaliberHighCaliber</iconPath>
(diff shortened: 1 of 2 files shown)


LockdownX7 (changes requested):

LockdownX7 (approved):

sumghai:
@Tostov - I noticed one small discrepancy between your XML def and the actual results from spreadsheet calculations.

I took the data you mentioned in your OP and crunched it through the spreadsheet:



Producing the following result:



The final mass comes out to be 0.17, but your `statBases/Mass` value in XML is 0.185; also, the calculated `MarketValue` should be 0.78 rather than the 0.91 in XML.

Finally, according to [this webpage](http://www.steyr-aug.com/amr.htm), the 150g value quoted is actually the complete cartridge mass (bullet, casing, propellant), so for CE, shouldn't the Bullet Case Mass really be `150 - 19.5=` 130.5 g?

PR #911, opened 2019-11-17
Xenohumans patch

zozilin:
Might not work until mod author removes the outdated patch from the mod itself.


Tostov:
Xenohumans have innate armor but use the human body type, so it won't apply. It may be worth giving them a custom bodytype identical to the human one with added CoveredByNaturalArmor groups on the appropriate body bits.

zozilin:
I don't really now how to do that.

PR #914, opened 2019-11-18
Add natural armor for humans

LockdownX7:
Adds CoveredByNaturalArmor to human outer body parts, allowing alien races using the human bodyDef to have sharp/blunt armor applied.

### DIFF
--- a/Patches/Core/Bodies/Bodies_Humanlike/Human.xml
+++ b/Patches/Core/Bodies/Bodies_Humanlike/Human.xml
@@ -199,13 +199,6 @@
   </Operation>
 
   <Operation Class="PatchOperationAdd">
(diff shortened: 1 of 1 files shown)


Patches/Core/Bodies/Bodies_Humanlike/Human.xml
```diff
@@ -175,5 +175,105 @@
...
+  </Operation>
+
+  <Operation Class="PatchOperationAdd">
+    <xpath>Defs/BodyDef[defName = "Human"]/corePart/parts/li[def="Neck"]/parts/li[def = "Head"]/groups</xpath>
+    <value>
+      <li>CoveredByNaturalArmor</li>
+    </value>
+  </Operation>
+
+  <Operation Class="PatchOperationAdd">
```
NoImageAvailable: Skull is an interior part and should not be covered

PR #917, opened 2019-11-20
Remove Slash

LockdownX7:


LockdownX7:
I included a merge from master so that the change would cover all integrated patches, these 2 commits are the only changes relevant to the PR:
d209604 
f13d3fb

LockdownX7:
If #924 is merged, I'll cover the recent patches in this PR as well.

PR #921, opened 2019-11-23
CE Patch for Kijin 2.0

Tostov:
Compatability patch for Kijin 2.0.

There's apparently texture bugs or something that occur, but I didn't run into any testing the stuff.


NoImageAvailable:
Is a separate ammo set actually warranted?

Tostov:
Both bows are pretty much designed around their abillity to apply their respective statuses, i.e. setting targets on fire  for the fire bow or the mod's poison hediff for the poison one.

Giving them custom ammo sets that addtionally apply their respective effects seemed like the simpler method, as compared to giving each a individual new ammo to have.

zozilin:
I get pink squares for any headwear on Kijin.

NoImageAvailable:
I'm talking about whether this ammo is specific enough to be race-specific. Flaming/poisoned arrows are generic enough they could show up in other mods.

Tostov:
That's a fair call. Should I add them to this PR, or should a new one be opened since they'd be non-specific to this patch?

Dyrewulfe:
I'm working on a fix for the rendering issues.

NoImageAvailable:
Just add them to this patch

Defs/Ammo/Neolithic/Arrows.xml
```diff
@@ -18,6 +18,8 @@
       <Ammo_Arrow_Stone>Projectile_Arrow_Stone</Ammo_Arrow_Stone>
       <Ammo_Arrow_Steel>Projectile_Arrow_Steel</Ammo_Arrow_Steel>
       <Ammo_Arrow_Plasteel>Projectile_Arrow_Plasteel</Ammo_Arrow_Plasteel>
+	  <Ammo_Arrow_Venom>Projectile_Arrow_Venom</Ammo_Arrow_Venom>
+	  <Ammo_Arrow_Plasteel>Projectile_Arrow_Flame</Ammo_Arrow_Plasteel>
```
LockdownX7: Ammo_Arrow_Plasteel should be Ammo_Arrow_Flame

AthenaSulisMinerva:
@Tostov - How were the crafting and damage stats for the Venom and Flame arrow ammo types calculated?

Tostov:
Venom matches steel but uses the venomarrow damage type (which works in the same manner as the snake bite damage type, though should have appropriate labels). Flame is half steel rounding down.

Crafting/Value for herbal matches FSX (for easy conversion on the ammo chart, and crafting the venom would take some time anyway, so it has the same processing value). Flame uses added prometium to recipe requirements on the table, and therefore increases the build time/value accordingly. PR has been made to standardize this properly.

AthenaSulisMinerva:
@Tostov - we have not been able to reproduce your values using the current CE 1.6/1.8+ projectiles spreadsheet:

https://docs.google.com/spreadsheets/d/1-y3KCuTM_vJnxw-gQPl5KIiz7ySd7fw9P6Wx4zB5bZg/edit#gid=1393347070

capture 01

capture 02

capture 03

capture 04

Could you show us the calculations from your spreadsheet?

PR #926, opened 2019-11-25
Apparel compatibility for BBBodySupport, move compatibility patches to subdir.

Dyrewulfe:
Adds apparel rendering compatibility for BBBodySupport, needed to support #921 

I took the soft touch with the logging, it's not an ancient DetoursLib job this time. And opinions on patching approaches aside, the prefix was written to detour for it's particular case only(which made it simple to patch).

Also moved this and the GraphicApparelDetour patch to a subdir and nested namespace, to avoid clutter.

Edit: Forgot to mention, this is currently dependent on author of BBBody removing a def file that overwrites the MiddleHead layer def(zozlin contacted the author). Seeing as that file is called "CE comp temp.xml", it's likely they'll oblige.

### DIFF
--- a/Source/CombatExtended/CombatExtended.csproj
+++ b/Source/CombatExtended/CombatExtended.csproj
@@ -261,13 +261,14 @@
     <Compile Include="CombatExtended\Verbs\Verb_MarkForArtilleryCE.cs" />
     <Compile Include="CombatExtended\Verbs\Verb_ShootMortarCE.cs" />
     <Compile Include="CombatExtended\Verbs\Verb_ShootCEOneUse.cs" />
+    <Compile Include="Harmony\Compatibility\Harmony-BBBodySupport.cs" />
     <Compile Include="Harmony\Harmony-BuildingProperties.cs" />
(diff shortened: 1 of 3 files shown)


ssulunge:
Hello this is BBBody author. Is removing def file all that I need to do? I mean I add that for temporal fix. and another temporal fix is written in the code as well. It checks if apparel is middlehead or not. If I were to delete def file, I need to delete that too.

Dyrewulfe:
@ssulunge Ah, yes, I see your issue.

The attached patch will cover that. The .txt extension will need to be removed, of course(Github doesn't think attaching source files is "appropriate", it seems.)

[CE Patch.xml.txt](https://github.com/NoImageAvailable/CombatExtended/files/3888035/CE.Patch.xml.txt)

ssulunge:
@Dyrewulfe
Thanks for pointing out that Zombiefied patch... that was my mistake bad copy of my code. It is fixed
Def file is deleted, and all middleHead stuffs were removed from code too. (from both BBbody mod and Kijin 1.0)

Please proceed updating BB patch on CE. and thanks for your hard work making all this patch.

PR #928, opened 2019-11-27
RunAndGun compatibility, prevent mobile reload attempts

### LINKED ISSUE #782: Unable to reload when fleeing (Run'n'Gun)
There is a mod called Run and Gun by roloo (https://steamcommunity.com/sharedfiles/filedetails/?id=1204108550). Probably, you know what it does - ability to shoot while moving. That issue is appear when using both Combat Extended and Run'n'Gun.
Enemies are able to reload and shoot when moving and attacking. Able to shoot when fleeing. But unable to reload when fleeing. They start infinite reload loop. When ammo depleted in weapon, they start to reload then immediately stops and starts to reload again. I can see "Reload" text spam on top of colonist.
I don't know what is the issue. But I think there is a check in Combat Extended that prevents successful reloading. Just want to ask for some kind of a fix.

Dyrewulfe:
This seems like the easiest way to fix this. Being able to reload some weapons while moving would be cool and all, but that would only be realistic for some weapons. And more work than I'm interesting in putting in.

This approach, at least, fixes the conflict.

Closes #782

### DIFF
--- a/Source/CombatExtended/CombatExtended/Comps/CompAmmoUser.cs
+++ b/Source/CombatExtended/CombatExtended/Comps/CompAmmoUser.cs
@@ -22,6 +22,7 @@ namespace CombatExtended
 
         public Building_TurretGunCE turret;         // Cross-linked from CE turret
 
+        internal static Type rgStance = null;       // RunAndGun compatibility, set in relevent patch if needed
         #endregion
 
         #region Properties
@@ -294,6 +295,10 @@ namespace CombatExtended
                 turret.TryOrderReload();
                 return;
             }
+
+            // R&G compatibility, prevents an initial attempt to reload while moving
+            if (Wielder.stances.curStance.GetType() == rgStance)
+                return;
 
             if (UseAmmo)
             {
--- a/Source/CombatExtended/Harmony/Compatibility/Harmony-RunAndGun.cs
+++ b/Source/CombatExtended/Harmony/Compatibility/Harmony-RunAndGun.cs
@@ -1,0 +1,98 @@
+﻿using Harmony;
+using RimWorld;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Verse;
+using System.Reflection.Emit;
+using System;
+
+
+namespace CombatExtended.Harmony.Compatibility
+{
+    [HarmonyPatch]
+    class Harmony_Compat_RunAndGun
+    {
+        static readonly string logPrefix = Assembly.GetExecutingAssembly().GetName().Name + " :: ";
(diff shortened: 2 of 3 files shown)


Dyrewulfe:
Hold this open for now. I think the changes can be isolated to the Harmony namespace.

Source/CombatExtended/Harmony/Compatibility/Harmony-RunAndGun.cs
```diff
@@ -0,0 +1,95 @@
...
+using System;
+
+
+namespace CombatExtended.Harmony.Compatibility
+{
+    [HarmonyPatch]
+    class Harmony_Compat_RunAndGun
+    {
+        static readonly string logPrefix = Assembly.GetExecutingAssembly().GetName().Name + " :: ";
+        static readonly Assembly ass = AppDomain.CurrentDomain.GetAssemblies().
```
NoImageAvailable: I approve of this variable naming

Source/CombatExtended/CombatExtended/Comps/CompAmmoUser.cs
```diff
@@ -295,6 +296,10 @@ public void TryStartReload()
                 return;
             }
 
+            // R&G compatibility, prevents an initial attempt to reload while moving
+            if (Wielder.stances.curStance.GetType() == rgStance)
```
NoImageAvailable: What exactly is the purpose behind rgStance, to check if the shooter is currently shooting while moving? If so it might be better to simply check current movement speed, which is how a number of other Run and Gun/GiddyUp compatibility features are done.

Dyrewulfe: Just testing for the pawn being in motion blocks manually issued reloads while moving, sadly.

NoImageAvailable: Don't manually ordered reloads stop the pawn already? I imagine this can be circumvented by applying the check to automatically triggered reloads only or else by stopping the pawn as part of the gizmo.

Source/CombatExtended/Harmony/Compatibility/Harmony-RunAndGun.cs
Dyrewulfe: It's one of my favorite things about writing these patches targeting unreferenced assemblies, honestly.

Source/CombatExtended/CombatExtended/Comps/CompAmmoUser.cs
NoImageAvailable: Conversation resolved via Discord

PR #972, opened 2020-01-20
Update to Neolithic, Shells, Rockets, Shotguns.

N7Huntsman:
- Crossbow bolts, arrows, sling stones updated to be current with sheet.
- Update 50mm Type 89 mortar.
- 20mm autocannon rounds updated, transferred from 'cannon' ammo folder to 'high caliber' based upon past and modern use (against contact vs area targets). 
- Updated all rockets and missiles to match sheet.


sumghai:
@N7Huntsman - For High Caliber vs Cannon shells, what caliber would you consider to be the cutoff diameter?

Personally, I think that 20, 25 and 30mm shells should be classified as High Caliber, while 40 and 57mm Bofors shells are large enough to be considered Cannon shells (since the latter are mainly used for AA or naval artillery).

N7Huntsman:
@sumghai Right around 30mm kind of ends up being a convenient cut-off point for a variety of reasons. However, the most straightforward rule of thumb I've been able to come up with is, "is there a rifle somewhere that fires this (even a homemade one; there are some 30mm 'sniper rifles' made from salvaged anti-air gun barrels that have popped among militants in Syria and Iraq)?" If yes, it's High Caliber. If not, it's a shell.

If a determining factor is _really_ needed beyond that, my next principle is whether or not the shell is designed to detonate _near_ or to detonate _against_.

Issue #937, opened 2019-12-09
[Fix in #1039] Reloading a weapon does not update its weight/bulk values.


Tostov:
When reloading a weapon, the ammo's mass/bulk that is loaded into the gun is not accounted for after the reload, until the inventory is altered again (e.g. picking up/dropping a item, shooting a weapon, equipping a different weapon.)

Tostov:
A simmilar issue occurs when unloading a weapon as well. The previously loaded ammo is added to the inventory, but its bulk/mass from being loaded in the gun remains until the inventory is altered again.

Alicecomma:
The issue suggests that there's a missing CompInventory.UpdateInventory call in CompAmmoUser.LoadAmmo:

LoadAmmo considers the case of ammo == null (unloading) without any calls to UpdateInventory. So, a fix would be to add "if (CompInventory != null) CompInventory.UpdateInventory();".
Similarly, LoadAmmo considers the case where all ammo in the inventory is used up without any calls to UpdateInventory.

Alicecomma:
Further inspection:
- Loading a weapon decreases inventory mass/bulk but doesn't change weapon mass/bulk
- Unloading increases mass/bulk but doesn't change weapon mass/bulk
- Dropping a weapon and equiping it again DOES check whether the weapon is loaded, adding mass/bulk for the loaded weapon: CompInventory.GetEquipmentStats considers loaded bullet weights.

----

An alternative way to fix this issue is to have a statworker for both mass and bulk, which takes into consideration loaded bullets. This way, loaded guns are heavier also in caravans. This is the desired change:

- Add StatWorker_Mass, which considers loaded rounds
- Add StatWorker_Bulk, which considers loaded rounds

----

The issue with such a change is that if there's other mods with custom StatWorker for Mass or Bulk, they are not accounted for properly. Perhaps there's an alternative way to update these stats on a Thing level.

Alicecomma:
Simplified the StatWorker stuff into a single StatWorker_LoadedAmmo, which considers Mass or Bulk separately for whatever stat (mass or bulk) it is attached to. This means that any StatDef could re-use this method of accessing loaded projectiles.

![image](https://user-images.githubusercontent.com/4662406/74784683-bf8e9f80-52a8-11ea-8f7e-c131f5c01c24.png)

Alicecomma:
Issue to be resolved:
- Add XML tag to weapons to decide whether loading ammo into them increases bulk/wornbulk

PR #999, opened 2020-02-18
Fix for #937

Alicecomma:
- CE_StatsReport_LoadedAmmo added to language keys (needs translation)
- Stats.xml patched to add CombatExtended.StatPart_LoadedAmmo to Mass stat
- Added file StatPart_LoadedAmmo
- CompAmmoUser now immediately calls UpdateInventory WHENEVER CurMagCount is changed, as it should be
- CompInventory stopped looking at loaded ammo for weight/bulk
- StatPart_LoadedAmmo was added, which does everything previously done in CompInventory, but which is common in current RimWorld versions


Alicecomma:
One additional conflict is introduced from CE:FT merger, which may be resolved after the merger.

Alicecomma:
Remade the PR in order to remove the included .dll

PR #1011, opened 2020-02-19
Set selectedAmmo to correct value on spawned pawns

Alicecomma:
Without this change, the ammo gizmo displays the default ammo on spawned pawns.


zozilin:
image
Doesn't seem to work

Alicecomma:
Yea, no .dll included

PR #1012, opened 2020-02-19
Fix for #460 (multi-selection of gun users allows ammo change for all)

### LINKED ISSUE #460: [Fix in #1012] Can't change ammo type of multiple select turrets simultaneously
Small UX issue, the reload gizmo doesn't override `InheritInteractionsFrom` from `Verse.Gizmo` which disallows players setting the ammo type of several emplacements at the same time.

I did some research into the code and it looks like it would require a restructure of the flow since `InheritInteractionsFrom` happens immediately on clicking of the gizmo before the user selects the ammo type from the drop down menu.

Alicecomma:
- GroupsWith & MergeWith allow combining reloading gizmos when multiple things with guns are selected. All are stored in list "others"
- BuildAmmoOptions is refactored to allow taking into account multiple selected gun users:
- - Switching ammo for multiple users shows "(1/2) AP", for example, indicating that 1 gun is loaded with AP rounds and that there is another gun which could be loaded with these rounds.

image

- - Reloading multiple guns is possible

image

- - Unloading multiple guns is possible

image

Commit overall fixes #460

### DIFF
--- a/Source/CombatExtended/CombatExtended/Gizmos/Command_Reload.cs
+++ b/Source/CombatExtended/CombatExtended/Gizmos/Command_Reload.cs
@@ -11,7 +11,27 @@ namespace CombatExtended
 {
     public class Command_Reload : Command_Action
     {
+        List<Command_Reload> others;
         public CompAmmoUser compAmmo;
+
+        public override bool GroupsWith(Gizmo other)
+        {
+            var order = other as Command_Reload;
+            return order != null;
+        }
+
+        public override void MergeWith(Gizmo other)
+        {
+            var order = other as Command_Reload;
+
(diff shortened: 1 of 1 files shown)


zozilin:
This is really neat!
Does it take into account ammo present in inventories or does it show that it could be loaded in principle? 
Will look into merging it and everything else after FT is done completely

Alicecomma:
It takes into account ammo in pawn inventories, but doesn't take into account for turrets. This is the same behaviour as previously, except now it can be applied to many ammo users at once.

Alicecomma:
Testing whether the change holds up with ammo system disabled

PR #1039, opened 2020-02-21
Fix #937 (loaded ammo has weight/bulk update properly)

### LINKED ISSUE #937: [Fix in #1039] Reloading a weapon does not update its weight/bulk values.

Alicecomma:
- CE_StatsReport_LoadedAmmo added to language keys (needs translation)
- Stats.xml patched to add CombatExtended.StatPart_LoadedAmmo to Mass stat
- Added file StatPart_LoadedAmmo
- CompAmmoUser now immediately calls UpdateInventory WHENEVER CurMagCount is changed, as it should be
- CompInventory stopped looking at loaded ammo for weight/bulk
- StatPart_LoadedAmmo was added, which does everything previously done in CompInventory, but which is common in current RimWorld versions
- Bulk StatDef now uses StatPart_LoadedAmmo
- AmmoUser CompProperties include a loadedAmmoBulkFactor with default value 0f. This allows RPG-7 and similar weapons to have a higher bulkiness (volume) when loaded.
- StatPart_LoadedAmmo multiplies by loadedAmmoBulkFactor when parentStat is Bulk

### DIFF
--- a/Defs/Stats/Stats_Basics_Inventory.xml
+++ b/Defs/Stats/Stats_Basics_Inventory.xml
@@ -10,6 +10,9 @@
     <minValue>0.001</minValue>
     <toStringStyle>FloatTwo</toStringStyle>
     <displayPriorityInCategory>10</displayPriorityInCategory>
+	<parts>
+		<li Class="CombatExtended.StatPart_LoadedAmmo"/>
+	</parts>
   </StatDef>
   
   <StatDef>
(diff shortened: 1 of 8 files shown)


Alicecomma:
Closed previous PR due to recent changes from CE:FT merger.

Alicecomma:
Testing with ammo system disabled, it could be that the CurrentAmmo reference is an issue

Alicecomma:
May be well-served with a StatDef instead of a CompProperties_AmmoUser for the loadedAmmoBulk stat. Do not merge until #1044 is resolved.

Issue #1038, opened 2020-02-21
[Fix in #1040] Fire, plasteel and toxic arrows spawn


Alicecomma:
Basically generateAllowChance = 0f doesn't do anything. Need to look at the code behind giving pawns ammo to see what does.

Alicecomma:
Harmony_PawnInventoryGenerator_GenerateInventoryFor
- LoadoutPropertiesExtension.GenerateLoadoutFor(Pawn pawn)
- - LoadoutPropertiesExtension.LoadWeaponWithRandAmmo(ThingWithComps gun)
- - - Selects from all ammoDefs in ammoSetDef, with alwaysHaulable
- - - Random element weighted by **ThingDef.generateCommonality**

Therefore, to disallow a type of ammo from being generated, set it to be alwaysHaulable = false (bad practice) or set generateCommonality = 0f.

Alicecomma:
Could be fixed:
- Introduce typical XML tags to disallow generation of ammo into C#

Or:
- Make sure everyone uses the correct tag.. somehow

Alicecomma:
Typical XML tags:
- ThingDef.generateAllowChance
- ThingDef.generateCommonality

Alicecomma:
Solution:
- Change LoadoutPropertiesExtension.LoadWeaponWithRandAmmo(ThingWithComps gun) to use generateAllowChance instead of generateCommonality.

PR #1040, opened 2020-02-21
Fix for #1038 (generateAllowChance instead of generateCommonality)

Alicecomma:
- 17 months ago, during RW version change, the code in the method was changed to generateCommonality from a completely different tag which doesn't exist anymore
- generateAllowChance is commonly used to prevent things from being generated, while generateCommonality is not for preventing, just for weighting
- All CE XML uses generateAllowChance, not generateCommonality
- Therefore, this code change is necessary


Alicecomma:
To test: Check that Arrows and GreatArrows do not spawn their Plasteel variant

Issue #1041, opened 2020-02-22
Warnings by AllYourBase tool


Alicecomma:
```
BaseBullet in file C:\Games\RimWorld2282Win64\Mods\CombatExtended\Defs\Ammo\AmmoBases.xml
MoteBase in file C:\Games\RimWorld2282Win64\Mods\CombatExtended\Defs\Effects\Mote_Visual.xml
TribalBase in file C:\Games\RimWorld2282Win64\Mods\CombatExtended\Defs\PawnKindDefs\PawnKinds_CE.xml
MercenaryBase in file C:\Games\RimWorld2282Win64\Mods\CombatExtended\Defs\PawnKindDefs\PawnKinds_CE.xml
BuildingBase in file C:\Games\RimWorld2282Win64\Mods\CombatExtended\Defs\ThingDefs_Buildings\Buildings_Defenses.xml
BenchBase in file C:\Games\RimWorld2282Win64\Mods\CombatExtended\Defs\ThingDefs_Buildings\Buildings_Production.xml
UnfinishedBase in file C:\Games\RimWorld2282Win64\Mods\CombatExtended\Defs\ThingDefs_Items\Items_Unfinished.xml
ApparelBase in file C:\Games\RimWorld2282Win64\Mods\CombatExtended\Defs\ThingDefs_Misc\Apparel_Various.xml
BaseFilth in file C:\Games\RimWorld2282Win64\Mods\CombatExtended\Defs\ThingDefs_Misc\Filth_Various.xml
BaseGrenadeProjectile in file C:\Games\RimWorld2282Win64\Mods\CombatExtended\Defs\ThingDefs_Misc\Weapons_Grenades.xml
BaseProjectileNeolithic in file C:\Games\RimWorld2282Win64\Mods\CombatExtended\Defs\ThingDefs_Misc\Weapons_RangedNeolithic.xml
BaseWeaponNeolithic in file C:\Games\RimWorld2282Win64\Mods\CombatExtended\Defs\ThingDefs_Misc\Weapons_RangedNeolithic.xml
PlantBase in file C:\Games\RimWorld2282Win64\Mods\CombatExtended\Defs\ThingDefs_Plants\Plants_Bases.xml
```

Alicecomma:
AmmoBases.xml:
- Why exactly is BaseBullet a Class="CombatExtended.AmmoDef". There is absolutely nothing that warrants this - all projectiles should be ThingDef, not AmmoDef
[...]

PR #1042, opened 2020-02-22
[XML-only] Fix for #1041 (removing bases, AllYourBase)

Alicecomma:


Alicecomma:
Should probably not be merged before discussing

N7Huntsman:
I'd suggest this waits until after the 1.1 release, then we can do some testing with large mod lists to ensure there aren't any serious compatibility issues, and if we're confident there aren't going to be issues and that it's what we want to do, we can add it to the release of the FT patches.

zozilin:
Reopen when 1.1 shenanigans are over

Issue #1037, opened 2020-02-21
M74 Incendiary rocket - adjust stackLimit or stack display graphics to depict real-life clip size


sumghai:
### Summary

The M74 Incendiary rockets are used by the M202 FLASH launcher, and the rockets are typically issued in an en bloc clip of four, rather than loaded individually into the launcher. 

Ideally, this should be reflected in the stack display of the ingame M74 ammo item.

### Current State

Currently, the M74 rockets in CE have a `stackLimit` of 25, and using the standard `Graphic_StackCount` graphicClass, we would see:

Stack size | Graphic displayed
--- | --- 
1|![M74_a.png](https://raw.githubusercontent.com/CombatExtendedRWMod/CombatExtended/Development/Textures/Things/Ammo/Rocket/M74/M74_a.png)
2~24|![M74_b.png](https://raw.githubusercontent.com/CombatExtendedRWMod/CombatExtended/Development/Textures/Things/Ammo/Rocket/M74/M74_b.png)
25|![M74_c.png](https://raw.githubusercontent.com/CombatExtendedRWMod/CombatExtended/Development/Textures/Things/Ammo/Rocket/M74/M74_c.png)

Whereas we should ideally have:

Stack size | Graphic displayed
--- | --- 
1|![M74_a.png](https://raw.githubusercontent.com/CombatExtendedRWMod/CombatExtended/Development/Textures/Things/Ammo/Rocket/M74/M74_a.png)
2~3|![M74_b.png](https://raw.githubusercontent.com/CombatExtendedRWMod/CombatExtended/Development/Textures/Things/Ammo/Rocket/M74/M74_b.png)
4|![M74_c.png](https://raw.githubusercontent.com/CombatExtendedRWMod/CombatExtended/Development/Textures/Things/Ammo/Rocket/M74/M74_c.png)

### Proposed Solutions
* **Option 1** - Simply set the `stackLimit` to 4
  * The main disadvantage is that this drastically reduces the amount of M74 rockets that can be stored per cell, and players would need more storage space even for a small stockpile of rockets
* **Options 2** - Set the `stackLimit` to a multiple of 4 (20 or 24), and use the `Graphic_StackCountRanged` graphicClass introduced from the FT merger to precisely control which graphic is displayed at what intermediate stack size (the ammo graphics need to be renamed slightly, since `Graphic_StackCountRanged` reads the number embedded in the filenames for the intermediate stack size)
[...]

PR #1043, opened 2020-02-22
Tweak M74 rocket stacklimit and stacked graphic display

sumghai:
For #1037 

Changes the display of stacked M74 rockets to better emulate the real-life appearance of 4 rockets in an complete en bloc clip:
* Reduce stackLimit to 24 (a multiple of 4)
* Change the graphicClass to the new `CombatExtended.Graphic_StackCountRanged` (introduced from the FastTrack merger)
* Rename the M74 ammo textures so that their filenames can be parsed by `Graphic_StackCountRanged` to display the appropriate texture at intermediate stack counts


zozilin:
I am okay with the graphic change. Stack limit change to 24 is very minor but it disrupts spreadsheet uniformity. 
What about a compromise - implement graphics, but keep stack limit as is?

sumghai:
Since the crafting recipe is currently set to 4 rockets per bill, if we keep the `stackLimit` to 25, we may have an odd situation where upon crafting the seventh iteration of the bill, one would end up with one full stack of 25 M74s, with 3 M74s left over.

Alicecomma:
That is a problem with the recipe system, and having a few rockets left-over is never really a problem

sumghai:
I have just reverted the `stackLimit` to 25, while keeping the graphics class changes.

PR #1045, opened 2020-02-22
Chinese translation by Cn-mjt44

zozilin:
@Cn-mjt44 that PR was still a bit messy, so I simply put everything here. Overall, looks good, thank you for your work


Cn-mjt44:
I see that, but it still has something not been add.So, can you add the all the thing inside?the old translation ar created for 3 years.

Cn-mjt44:
@zozilin

zozilin:
Added in #1048

PR #1046, opened 2020-02-22
Chinese translate remind

Cn-mjt44:
I found that still has something not been add. pelase take them all.


Cn-mjt44:
Languages/ChineseSimplified/DefInjected/ThingDef/Ammo_All.xml are not empty

zozilin:
I believe this is all added in #1045 and #1048

PR #1047, opened 2020-02-22
Update Xenn patches

sumghai:
* Reimplement Xenn ammo as conditional patches
* Update Xenn weapon melee tools
* Update Xenn race melee tools

Note - further revision of the Xenn ammo will be done post-CE 2.0:
* The two weapons are meant to fire different ammo (one hot, one cold), rather than sharing one universal ammoset
* The previous patch author incorrectly described a freezing ammo cartridge as thermobaric

### DIFF
--- a/Patches/Xenn/ThingDefs_Misc/Xenn_Weps.xml
+++ b/Patches/Xenn/ThingDefs_Misc/Xenn_Weps.xml
@@ -54,33 +54,30 @@
 							<capacities>
 								<li>Blunt</li>
 							</capacities>
-							<power>9</power>
-							<cooldownTime>1.8</cooldownTime>
-							<chanceFactor>1.5</chanceFactor>
-							<armorPenetration>0.11</armorPenetration>
+							<power>8</power>
+							<cooldownTime>1.79</cooldownTime>
 							<linkedBodyPartsGroup>Stock</linkedBodyPartsGroup>
+							<armorPenetrationBlunt>2.755</armorPenetrationBlunt>
 						</li>
 						<li Class="CombatExtended.ToolCE">
-							
 							<label>barrel</label>
 							<capacities>
 								<li>Blunt</li>
 							</capacities>
-							<power>10</power>
-							<cooldownTime>1.9</cooldownTime>
-							<armorPenetration>0.118</armorPenetration>
+							<power>5</power>
+							<cooldownTime>3.02</cooldownTime>
 							<linkedBodyPartsGroup>Barrel</linkedBodyPartsGroup>
+							<armorPenetrationBlunt>1.63</armorPenetrationBlunt>
 						</li>
 						<li Class="CombatExtended.ToolCE">
-							
 							<label>muzzle</label>
 							<capacities>
 								<li>Poke</li>
 							</capacities>
-							<power>10</power>
-							<cooldownTime>1.9</cooldownTime>
-							<armorPenetration>0.086</armorPenetration>
+							<power>8</power>
+							<cooldownTime>1.55</cooldownTime>
 							<linkedBodyPartsGroup>Muzzle</linkedBodyPartsGroup>
+							<armorPenetrationBlunt>2.755</armorPenetrationBlunt>
 						</li>
 					</tools>
 				</value>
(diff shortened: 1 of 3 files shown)


N7Huntsman:
> The two weapons are meant to fire different ammo (one hot, one cold), rather than sharing one universal ammoset

I'm not entirely certain what you mean, but I'm thinking it shouldn't be too difficult to configure once the ammo has been made.

Patches/Xenn/ThingDefs_Races/Race_Xenn.xml
```diff
@@ -4,121 +4,125 @@
...
-		<value>
-			<li>
-				<compClass>CombatExtended.CompPawnGizmo</compClass>
+			<!-- ========== Turn our dog boys into dog *boys* ========== -->
+			<!-- ========== To be human is to be a tall brick ========== -->
+			<li Class="PatchOperationAddModExtension">
+				<!-- === Note to future furball compatibility nerds: 		=== -->
+				<!-- === Paste the defName of your horrible alien there   |	=== -->
+				<!-- === 												  V	=== -->
+				<xpath>*/AlienRace.ThingDef_AlienRace[defName="Alien_Xenn"]</xpath>
```
N7Huntsman: In all cases, please replace `<xpath>*/ ...` with `<xpath>/Defs/ ...` as  */ loads far slower.  `<xpath>Defs/ ...` is also acceptable to be left in place.

Patches/Xenn/ThingDefs_Misc/Xenn_Weps.xml
```diff
@@ -54,33 +54,30 @@
 							<capacities>
 								<li>Blunt</li>
```
N7Huntsman: It's no longer necessary for ranged weapons to have individualized melee attack tools calculated. As this appears to be configured like a standard, non-pistol, non-heavy weapon firearm, you can just copy-paste the melee attack stats CE gives the various shotguns and rifles in `.../Patches/Core/ThingDefs_Misc/Weapons_Guns`.

Patches/Xenn/ThingDefs_Races/Race_Xenn.xml
```diff
@@ -4,121 +4,125 @@
 	<Operation Class="PatchOperationSequence">
 		<success>Always</success>
 		<operations>
-	<!-- ========== Look for Fido ========== -->
-	<li Class="CombatExtended.PatchOperationFindMod">
+			<!-- ========== Look for Fido ========== -->
+			<li Class="CombatExtended.PatchOperationFindMod">
```
N7Huntsman: I strongly suspect this is just a leftover from an old patch rather than FT's work, but as a general remark, `CombatExtended.PatchOperationFindMod` is no longer used in favor of the vanilla `PatchOperationFindMod`. While I won't necessitate this be changed for this (or any incoming FT patch) b/c of the tedious and pointless .xml work, newer patches should be written using the vanilla patch op.

sumghai:
>The two weapons are meant to fire different ammo (one hot, one cold), rather than sharing one universal ammoset
>> I'm not entirely certain what you mean, but I'm thinking it shouldn't be too difficult to configure once the ammo has been made.

The original Xenn Race mod included two ranged weapons
* Xenn burn cannon, which only fires a plasma/napalm bolt (descriptions are inconsistent)
* Xenn frost cannon, which only fires a cyrogenic "freeze" bolt

The _correct_ way to have implemented ammo for these weapons would be separate ammosets:
```
<CombatExtended.AmmoSetDef>
    <defName>AmmoSet_XennFuel_BurnCannon</defName>
        <label>Xenn Fuel Cell</label>
            <ammoTypes>
                <Ammo_Xenn_Hot>Bullet_Xenn_Hot</Ammo_Xenn_Hot>
            </ammoTypes>
</CombatExtended.AmmoSetDef>

<CombatExtended.AmmoSetDef>
    <defName>AmmoSet_XennFuel_FrostCannon</defName>
        <label>Xenn Fuel Cell</label>
            <ammoTypes>
                <Ammo_Xenn_Cold>Bullet_Xenn_Cold</Ammo_Xenn_Cold>
            </ammoTypes>
</CombatExtended.AmmoSetDef>
```

However, the person who wrote the previous version of the ammo def/patch basically put them all in the same ammoset, which means both weapons can interchangeably use each other's ammo, in contradiction to the original intent of the Xenn mod:

```
<CombatExtended.AmmoSetDef>
    <defName>AmmoSet_XennFuel</defName>
        <label>Xenn Fuel Cell</label>
            <ammoTypes>
                <Ammo_Xenn_Hot>Bullet_Xenn_Hot</Ammo_Xenn_Hot>
                <Ammo_Xenn_Cold>Bullet_Xenn_Cold</Ammo_Xenn_Cold>
                <Ammo_Xenn_Foam>Bullet_Xenn_Foam</Ammo_Xenn_Foam>
            </ammoTypes>
</CombatExtended.AmmoSetDef>
```

One of the comments in Xenn_Weps.xml indicates this was not meant to be final:
> \
> \

For now, I'm going to keep the current implementation, and revamp the ammo when I have more time.

Patches/Xenn/ThingDefs_Races/Race_Xenn.xml
sumghai: This particular instance was indeed left over from an old patch.

And yeah, quite a few actual FT-exclusive patches also used this PatchOp, but I'll remember to use the vanilla `PatchOperationFindMod` from now on.

Patches/Xenn/ThingDefs_Races/Race_Xenn.xml
sumghai: I'll change those to `Defs/` ASAP. I've never really liked `*/` anyway, for the exact reason you stated.

sumghai: Fixed in ce3d2f3

Patches/Xenn/ThingDefs_Misc/Xenn_Weps.xml
sumghai: Fixed in 615aa79

PR #1048, opened 2020-02-22
Chinese translation by Cn-mjt44 part 2

zozilin:


Cn-mjt44:
I think you just need to put the patch-18.because, the patch-18 hase been fully translated.

PR #1053, opened 2020-02-23
Update More Mechanoids patches

sumghai:

* Change Assaulter weapon to use 6x18mm Charged, instead of 5x35mm Charged
  * As justified by @NickZiggy's in his original commit to FT:
Annotation 2020-02-23 111844

### DIFF
--- a/Patches/MoreMechanoids/ThingDefs_Misc/Weapons_Mechanoid.xml
+++ b/Patches/MoreMechanoids/ThingDefs_Misc/Weapons_Mechanoid.xml
@@ -21,19 +21,19 @@
 				  <recoilAmount>1.28</recoilAmount>
 				  <verbClass>CombatExtended.Verb_ShootCE</verbClass>
 				  <hasStandardCommand>true</hasStandardCommand>
-				  <defaultProjectile>Bullet_6x18mmCharged</defaultProjectile>
+				  <defaultProjectile>Bullet_5x35mmCharged</defaultProjectile>
 				  <warmupTime>1.1</warmupTime>
 				  <range>86</range>
 				  <ticksBetweenBurstShots>3</ticksBetweenBurstShots>
 				  <burstShotCount>80</burstShotCount>
-				  <soundCast>Shot_ChargeRifle</soundCast>
-				  <soundCastTail>GunTail_Light</soundCastTail>
+				  <soundCast>ChargeLance_Fire</soundCast>
+				  <soundCastTail>GunTail_Heavy</soundCastTail>
 				  <muzzleFlashScale>9</muzzleFlashScale>
 				</Properties>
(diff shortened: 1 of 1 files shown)


Tostov:
Personal taste, but maybe it'd be more prudent to adjust the range/burst amount rather than specifically the ammo type. If we are to resolve switching the ammo type, it may also be worth reviewing the weapon's recoil values, since the 6x18 should have a fair bit less recoil than the 5x35 does.

sumghai:
@NickZiggy, could you elaborate further on the rationale for the change to @Tostov?

N7Huntsman:
The 5x35mm charged round is effectively an anti-material round, so I really think it isn't intended for a automatic weapon. At 30 Sharp AP, it's going to go clean through basically anything and everything. I'm happy enough with swapping for 6x18mm in this case. If it's felt it needs to be a true heavy weapon, then use 12x64mm like the existing Heavy Charge Blaster.

Tostov:
The thing to keep in mind is that the 5x35 isn't a anti-material so much as a high penetration round. While it does go clean through any non-power armor setups (high quality power armor and power armor paired with a ballistic shield can both stop 5x35, and with its lack of blunt pen it will tend to hard block it rather than just reduce the damage) its actual overall damage output isn't particuarly high. 

The reason why this high penetration is important, is because the assaulter is more of a suppression platform rather than a closer quarters brawler. It fires from very long range with a weapon that has very little chance of securing actual hits outside of point blank range thanks to its own absurd rate of fire, burst length and lack of optics. Its high penetraiton however, allows it to supress most targets even if they're wearing power armor, since the high penetration effect cancels out the armor coverage. By changing the ammo to 6x18mm, its loses this role since the lack of penetration of the round prevents it from effectively supressing heavily armored targets.

The Assaulter may have good sharp armor, but its reletavely low blunt armor and general lack of HP as compared to the centipede or other heavy hitters doesn't make it a specialized close quarters unit (even if the gun, like every mechanoid gun except the charge lance, is deadly in cqc due to extensive volume of fire). Addtionally unlike the scyther or centipede, it isn't particuarly deadly in melee proper.

The other issue with exchanging the 5x35 for 6x18, is that it allows a player who defeats one assaulter to functionally ignore the charge ammo restriction/costs for a charge SMG, because the things are loaded with a few thousand thanks to the extesnsive mag size.

The final reason I largely am against the specific change is that the original reason for changing it (its one hit abillity) is largely incorrect. The weapon is incapable of one hit kills on anything but a brain hit, and most weapons with sufficent penetration to punch through a helmet will tend to do that.

sumghai:
Fair enough, @Tostov.

And seeing as @NickZiggy never replied to our request for additional info/justification, I'm going to revert this change.

PR #1055, opened 2020-02-23
Update FashionRIMsta patches

sumghai:

* Add `SmokeSensitivity` equippedStatOffset, `WearingGasMask` hediff and `StrappedHead` apparel layer to gas mask
* Fix indentation

### DIFF
--- a/Patches/FashionRIMsta/ThingsDefs_Misc/FashionRIMsta_Apparel.xml
+++ b/Patches/FashionRIMsta/ThingsDefs_Misc/FashionRIMsta_Apparel.xml
@@ -1,163 +1,186 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Patch>
 
-  <Operation Class="PatchOperationSequence">
-    <success>Always</success>
-    <operations>
-      <li Class="CombatExtended.PatchOperationFindMod">
-        <modName>FashionRIMsta</modName>
-      </li>
+	<Operation Class="PatchOperationSequence">
+		<success>Always</success>
+		<operations>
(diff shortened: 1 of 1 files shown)


Patches/FashionRIMsta/ThingsDefs_Misc/FashionRIMsta_Apparel.xml
```diff
@@ -1,163 +1,186 @@
...
+			</li>
+
+			<li Class="PatchOperationRemove">
+				<xpath>Defs/ThingDef[defName="Apparel_FSOveralls"]/equippedStatOffsets/WorkSpeedGlobal</xpath>
+			</li>
+
+			<li Class="PatchOperationAdd">
+				<xpath>Defs/ThingDef[defName="Apparel_FSOveralls"]/equippedStatOffsets</xpath>
+				<value>
+					<CarryBulk>10</CarryBulk>
```
N7Huntsman: As basic clothing (shirt, pants, etc.) in CE doesn't add CarryBulk, the clothing items here may be somewhat too high--could perhaps get away with up to maybe 5 at most for especially spacious items?

sumghai: These values were originally defined in 2017 here in CE, and the FT team made no further adjustments since.

That said, I can cap most of these down to `<CarryBulk>5</CarryBulk>`.

sumghai: Fixed in f30462c

PR #1056, opened 2020-02-23
Update DMGT pack patches

sumghai:

* Turret weapon stats re-calculated using the [CE:FT Gun Stats spreadsheet](https://docs.google.com/spreadsheets/d/1Z9ZBnwRQWCKQm18O4I9ja4mFwLVChkNNfam2ZeQTTa0/edit#gid=1573763037)
* Reduced a number of turret ranges to more reasonable values
* Fixed incorrect turret range display radius

### DIFF
--- a/Patches/DefensiveMachineGunTurretPack/ThingDefs/DMGT_Defensive_Turrets.xml
+++ b/Patches/DefensiveMachineGunTurretPack/ThingDefs/DMGT_Defensive_Turrets.xml
@@ -12,18 +12,18 @@
 			<li Class="CombatExtended.PatchOperationMakeGunCECompatible">
 				<defName>Gun_DP28MG</defName>
 				<statBases>
-					<RangedWeapon_Cooldown>0.56</RangedWeapon_Cooldown>
-					<SightsEfficiency>1</SightsEfficiency>
+					<RangedWeapon_Cooldown>0.35</RangedWeapon_Cooldown>
+					<SightsEfficiency>1.00</SightsEfficiency>
 					<ShotSpread>0.05</ShotSpread>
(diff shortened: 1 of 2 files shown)


Patches/DefensiveMachineGunTurretPack/ThingDefs/DMGT_Defensive_Turrets.xml
```diff
@@ -162,17 +180,16 @@
 						<canTargetLocations>true</canTargetLocations>
 					</targetParams>
 					<muzzleFlashScale>10</muzzleFlashScale>
-					<recoilPattern>Mounted</recoilPattern>
```
N7Huntsman: recoilPattern removed?

sumghai: I was under the assumption that CE firearms no longer needed `recoilPattern`. Or do turrets still use these?

N7Huntsman: All I can tell you for certain is that _all_ of the turrets within CE still use `<recoilPattern>Mounted</recoilPattern>`, as does the LMG (which I presume is to essentially mean it's being fired from a bipod). Now, that doesn't guarantee it still _does anything,_ but it is still there.

As for other firearms, I can only assume there's some other, default value it defaults to when not specified that covers basically all other firearms.

sumghai: Fixed in 1f3f441

PR #1060, opened 2020-02-24
Make a patch for RimTasers Reloaded?

theoutlawvv:
Any way you could make a patch for RimTasers Reloaded? https://steamcommunity.com/sharedfiles/filedetails/?id=1745038195
I know the bean ammo does virtually the same thing, the issue is it only works if you use the ammo system. And i highly dislike the ammo system so i have it disable. I'd be happy to give you guys some money for your efforts if you could make this patch, or update CEFT so you can set a shotguns bullets to bean bullets without enabling the ammo system.


N7Huntsman:
You've created a Pull Request (a request to add files you already have to the game), rather than creating an Issue, which I can only assume is what you intended. However, we don't generally accept requests for patches via the Github. You can, however, potentially join the Discord to find some more resources for patching.

PR #1076, opened 2020-03-01
Update to non-weapon Android Tier patches

Tostov:
The latest update to android tiers renamed the mod in the about file. This adds that new name to the valid modnames for the patches.

Also some tweaks to the K9 robot dog and the advanced jaw, since the latter cannot be used by humanoid androids like I previously thought.

### DIFF
--- a/Patches/Android Tiers/Hediffs/Hediffs_Android.xml
+++ b/Patches/Android Tiers/Hediffs/Hediffs_Android.xml
@@ -145,7 +145,7 @@
 					</capacities>
 					<power>44</power>
 					<cooldownTime>2.51</cooldownTime>
-					<armorPenetrationSharp>14.07</armorPenetrationSharp>
+					<armorPenetrationSharp>9.38</armorPenetrationSharp>
 					<armorPenetrationBlunt>45</armorPenetrationBlunt>
 				</li>
 				</tools>
--- a/Patches/Android Tiers/Hediffs/Hediffs_Android.xml
+++ b/Patches/Android Tiers/Hediffs/Hediffs_Android.xml
@@ -4,6 +4,7 @@
 	<Operation Class="PatchOperationFindMod">
     <mods>
         <li>[1.0] Android tiers</li>
+		<li>Android tiers</li>
     </mods>
 	<match Class="PatchOperationSequence">
 		<operations>
@@ -142,10 +143,10 @@
 					<capacities>
 						<li>Bite</li>
 					</capacities>
-					<power>31</power>
-					<cooldownTime>1.86</cooldownTime>
-					<armorPenetrationSharp>4.22</armorPenetrationSharp>
-					<armorPenetrationBlunt>13.5</armorPenetrationBlunt>
+					<power>44</power>
+					<cooldownTime>2.51</cooldownTime>
+					<armorPenetrationSharp>14.07</armorPenetrationSharp>
+					<armorPenetrationBlunt>45</armorPenetrationBlunt>
 				</li>
 				</tools>
 			</value>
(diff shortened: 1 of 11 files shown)


Patches/Android Tiers/Hediffs/Hediffs_Android.xml
```diff
@@ -142,10 +143,10 @@
 					<capacities>
 						<li>Bite</li>
 					</capacities>
-					<power>31</power>
-					<cooldownTime>1.86</cooldownTime>
-					<armorPenetrationSharp>4.22</armorPenetrationSharp>
-					<armorPenetrationBlunt>13.5</armorPenetrationBlunt>
+					<power>44</power>
+					<cooldownTime>2.51</cooldownTime>
+					<armorPenetrationSharp>14.07</armorPenetrationSharp>
```
N7Huntsman: What creatures are these jaws for? They're extremely strong (for comparison, the Grizzly bear has the strongest vanilla Bite with 31 damage) and the Sharp AP may be too high.

Patches/Android Tiers/ThingDef_Races/Races.xml
```diff
@@ -407,11 +408,11 @@
 						<capacities>
 						  <li>Bite</li>
 						</capacities>
-						<power>22</power>
-						<cooldownTime>2.22</cooldownTime>
+						<power>26</power>
+						<cooldownTime>2.46</cooldownTime>
 						<linkedBodyPartsGroup>Teeth</linkedBodyPartsGroup>
```
N7Huntsman: Depending on how exactly you envision the Android dog, the bite may be a little too powerful. Depending on what you're going for, you might compare to the husky, wolf, and warg as points of reference.

Tostov:
First note, the cybernetic jaw is the upgraded jaw for the K9.

As for the statistics, the general idea is that the jaw for the K9 isn't really so much a regular animal jaw so much as a pnumatic pair of scissors shaped into a jaw like facade. For the regular jaw, this is more suble and is mostly just a case of it being somewhat more powerful pressure wise and and having a slower, sharper attack (since you're looking at machined blades rather than naturally formed/hardened teeth).

The cybernetic jaw however, is keyed more off of the scyther's penetration blade rather than a regular attack, representing a strictly weapon mounted in the mouth rather than a convetional "jaw" of any description. The 40+ damage is largely to get it to the point where it can reliably scissor off a leg, which is reasonably what I'd expect it to be able to do in this instance. I do think it is reasonable to lower the sharp penetration value, since it's conceptually a tool that works more off of the sheer amount of pressure rather than specifically the sharpness of the cutting edge.

The key driver behind these decisions is the sheer value of the K9 unit, it is:
A) Extremely expensive value wise, and fairly expensive cost wise, it takes 15 components + 100 plasteel and then some uranium, but is notably valued at 8.5k (notably the other attack animals, the bear and warg are 445 and 700 respectively, meaning it's between 10-20x as expensive)
B) Installing the jaw requires another 40 steel, 50 plasteel and 8 components, and slaps another 2k onto the K9's cost.

Patches/Android Tiers/Hediffs/Hediffs_Android.xml
```diff
@@ -145,7 +145,7 @@
 					</capacities>
 					<power>44</power>
 					<cooldownTime>2.51</cooldownTime>
-					<armorPenetrationSharp>14.07</armorPenetrationSharp>
+					<armorPenetrationSharp>9.38</armorPenetrationSharp>
 					<armorPenetrationBlunt>45</armorPenetrationBlunt>
```
N7Huntsman: The new sharp value is acceptable. The blunt AP seems high for a bite, but I presume it's to get through power armor?

Tostov: It is pretty much strictly to let it get through power armor, yeah.

N7Huntsman (approved):
With the adjusted stats, it works for me.

zozilin:
I'll have to revert it. Please reopen when 1.1 is done @Tostov

PR #1082, opened 2020-03-06
Development 1.0.2.0

zozilin:


sumghai:
@zozilin - would it be kosher if I submitted a PR for an updated preview image? I think it would be nice to mark the CE 2.0 milestone.

Essentially, it would be the current FastTrack image, without the "FastTrack Edition" text.

PR #1123, opened 2020-03-23
lower case mech ammo beacon

jtgibson01:
Simple hotfix of newly-added mech ammo beacon building label.


jtgibson01:
Sorry, this was meant to be a single-file commit to the dev branch.

PR #1150, opened 2020-03-27
Fgo styles1.1 patch update

Tostov:
## Additions

Added patches for new apparels added.

## Changes

All of the proper "Armor" type apparels have had their defense stats increased/adjusted.

## Reasoning

Previously in the parent mod the layers for the armor apparels were all over the place and it was kind of difficult to give them a concrete value. Now all occupy middle meaning they can't be stacked with an armor vest/power armor, so it's more reasonable to give them a comparable value. Addtionally, all the proper armor type ones have been limited behind the power armor tech research, so they're significantly less accessable early on.


N7Huntsman (approved):
Looks good.

N7Huntsman (commented):
I'm seeing some pretty odd behavior with the apparel in that certain clothing won't appear on certain pawn body shapes. I can only assume the base mod doesn't have a texture for them, or something? The armor and bulk stats you've patched in seem fine.

Schatten5 (approved):

Issue #1068, opened 2020-02-26
Import patches from FastTrack fork


sumghai:
A number of compatibility patches have already been brought over from the CE: FastTrack Edition fork, so this issue is primarily a TODO list of the remaining ones I (@sumghai) will be responsible for.

As always, PRs for each set of patches will undergo standard CE vetting.

* [ ] Alpha Animals
* [ ] Android Tiers
* [ ] AnimalCollabProj
* [ ] [CP] British Military Kit
* [ ] [CP] DOOM (1.0) 
* [ ] [CP] The DOOM Kit - Classic (1.0)
* [ ] [CP] Metal Gear Solid (1.0)
* [ ] [CP] Red Horse Furniture 
* [ ] [CP] Rimmu-Nation - Clothing
* [ ] [CP] Rimmu-Nation - Clothing (All Craftable!) (1.0) 
* [ ] [CP] Rimmu-Nation - Weapons
* [ ] [CP] Spec Ops - The Line (1.0)
* [ ] Flommel
* [ ] Forsakens
* [ ] Glitter Tech
* [ ] Kenshi
* [ ] Let's Have a Cat!
* [ ] Magical Menagerie
* [ ] More Vanilla Turrets
* [ ] NewRatkinPlus
* [ ] [Ods] Great War Imperial Germany - Uniforms (replaces [Ods] Kaiser Reich's Imperial Army - Apparel)
* [ ] [Ods] Third Reich - Uniforms
* [ ] [Pratt] WWII Weapons Pack (Vanilla)
* [ ] T-45b Power Armor
* [ ] [RH] Faction: Cordis Die (1.0)
* [ ] [RH] Faction: Elite Crew
* [ ] [RH] Faction: Last Man Contingent
* [ ] [RH] Faction: Militaires Sans Frontieres
* [ ] [RH] Faction: Task Force 141
* [ ] [RH] Faction: The Ghosts
* [ ] [RH] Faction: Umbra Company
* [ ] Rimworld of Magic
  * Major undertaking, due to many new bugs reported since the transition from CE 1.5 to 1.6
* [ ] Sergal, Gnolls and Lions
  * Partial implementation exists in CE; check what changes need to be merged
* [ ] Soviet Armory
* [ ] Spidercamp's Dog Pack
* [ ] Star Wars - Factions
  * Did not originally make it into an FT update prior to CE merger
* [ ] Warhammer: Gor

N7Huntsman:
[...]

PR #1151, opened 2020-03-27
Add AnimalCollabProj patches

sumghai:
## Additions

* Add compatibility patches for AnimalCollabProj, as part of the FastTrack merger.

## References

* Contributes towards #1068

## Reasoning

* Race stats taken from a [third-party standalone patch mod (that was then integrated into CE:FT FT-002)](https://steamcommunity.com/sharedfiles/filedetails/?id=1703449478)
* Animal melee tool sharp & blunt armor penetration stats rebalanced to 1.1 standards, based on CE stats for Core vanilla animals

## Testing

- [x] Compiles without warnings
- [x] Playtested a colony (1 hour)
  * Spawned two of each animal via dev menu
  * Verified that animal melee tools function properly by triggering a Psychic animal pulser and having the animals attack vanilla Human Thrasher and Mechanoid Scyther pawns

### DIFF
--- a/Patches/AnimalCollabProj/AnimalCollabProj_CE_Patch_Races_Animal_Bison.xml
+++ b/Patches/AnimalCollabProj/AnimalCollabProj_CE_Patch_Races_Animal_Bison.xml
@@ -1,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<Patch>
+	<Operation Class="PatchOperationSequence">
+		<success>Always</success>
+		<operations>
+
+			<li Class="CombatExtended.PatchOperationFindMod">
+				<modName>AnimalCollabProj</modName>
+			</li>
+
+			<li Class="PatchOperationAdd">
+				<xpath>Defs/ThingDef[defName="ACPBison"]/statBases</xpath>
+				<value>
+					<MeleeDodgeChance>0.08</MeleeDodgeChance>
+					<MeleeCritChance>0.18</MeleeCritChance>
+					<MeleeParryChance>0.15</MeleeParryChance>
+				</value>
+			</li>
+
+			<li Class="PatchOperationReplace">
+				<xpath>Defs/ThingDef[defName="ACPBison"]/tools</xpath>
+				<value>
+					<tools>
+						<li Class="CombatExtended.ToolCE">
+							<label>head</label>
+							<capacities>
+								<li>Blunt</li>
+							</capacities>
+							<power>6</power>
+							<cooldownTime>2.12</cooldownTime>
+							<chanceFactor>0.2</chanceFactor>
+							<linkedBodyPartsGroup>HeadAttackTool</linkedBodyPartsGroup>
+							<armorPenetrationBlunt>2</armorPenetrationBlunt>
+						</li>
+					</tools>
+				</value>
+			</li>
+
+		</operations>
+	</Operation>
+</Patch>
--- a/Patches/AnimalCollabProj/AnimalCollabProj_CE_Patch_Races_Animal_Bison.xml
+++ b/Patches/AnimalCollabProj/AnimalCollabProj_CE_Patch_Races_Animal_Bison.xml
@@ -11,10 +11,16 @@
 			<li Class="PatchOperationAdd">
 				<xpath>Defs/ThingDef[defName="ACPBison"]/statBases</xpath>
 				<value>
-					<MoveSpeed>5.6</MoveSpeed>
 					<MeleeDodgeChance>0.11</MeleeDodgeChance>
 					<MeleeCritChance>0.37</MeleeCritChance>
 					<MeleeParryChance>0.22</MeleeParryChance>
+				</value>
+			</li>
+
+			<li Class="PatchOperationReplace">
+				<xpath>Defs/ThingDef[defName="ACPBison"]/statBases/MoveSpeed</xpath>
+				<value>
+					<MoveSpeed>5.6</MoveSpeed>
 				</value>
 			</li>
 
--- a/Patches/AnimalCollabProj/AnimalCollabProj_CE_Patch_Races_Animal_Bison.xml
+++ b/Patches/AnimalCollabProj/AnimalCollabProj_CE_Patch_Races_Animal_Bison.xml
@@ -11,9 +11,10 @@
 			<li Class="PatchOperationAdd">
 				<xpath>Defs/ThingDef[defName="ACPBison"]/statBases</xpath>
 				<value>
-					<MeleeDodgeChance>0.08</MeleeDodgeChance>
-					<MeleeCritChance>0.18</MeleeCritChance>
-					<MeleeParryChance>0.15</MeleeParryChance>
+					<MoveSpeed>5.6</MoveSpeed>
+					<MeleeDodgeChance>0.11</MeleeDodgeChance>
+					<MeleeCritChance>0.37</MeleeCritChance>
+					<MeleeParryChance>0.22</MeleeParryChance>
 				</value>
 			</li>
 
@@ -24,13 +25,14 @@
 						<li Class="CombatExtended.ToolCE">
 							<label>head</label>
 							<capacities>
-								<li>Blunt</li>
+								<li>Stab</li>
 							</capacities>
-							<power>6</power>
-							<cooldownTime>2.12</cooldownTime>
-							<chanceFactor>0.2</chanceFactor>
+							<power>20</power>
+							<cooldownTime>3.09</cooldownTime>
 							<linkedBodyPartsGroup>HeadAttackTool</linkedBodyPartsGroup>
-							<armorPenetrationBlunt>2</armorPenetrationBlunt>
+							<chanceFactor>0.2</chanceFactor>
+							<armorPenetrationSharp>0.12</armorPenetrationSharp>
+							<armorPenetrationBlunt>9</armorPenetrationBlunt>
 						</li>
 					</tools>
 				</value>
--- a/Patches/AnimalCollabProj/AnimalCollabProj_CE_Patch_Races_Animal_DuckBilledPlatypus.xml
+++ b/Patches/AnimalCollabProj/AnimalCollabProj_CE_Patch_Races_Animal_DuckBilledPlatypus.xml
@@ -1,0 +1,63 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<Patch>
+	<Operation Class="PatchOperationSequence">
+		<success>Always</success>
+		<operations>
+
+			<li Class="CombatExtended.PatchOperationFindMod">
+				<modName>AnimalCollabProj</modName>
+			</li>
+
+			<li Class="PatchOperationAdd">
+				<xpath>Defs/ThingDef[defName="ACPDuckBilledPlatypus"]/statBases</xpath>
+				<value>
+					<MeleeDodgeChance>0.07</MeleeDodgeChance>
+					<MeleeCritChance>0.11</MeleeCritChance>
+					<MeleeParryChance>0.02</MeleeParryChance>
+				</value>
+			</li>
+
+			<li Class="PatchOperationReplace">
+				<xpath>Defs/ThingDef[defName="ACPDuckBilledPlatypus"]/tools</xpath>
+				<value>
+					<tools>
+						<li Class="CombatExtended.ToolCE">
+							<label>spurs</label>
+							<capacities>
+								<li>PlatypusSting</li>
+							</capacities>
+							<power>7</power>
+							<cooldownTime>1.9</cooldownTime>
+							<linkedBodyPartsGroup>Spurs</linkedBodyPartsGroup>
+							<armorPenetrationSharp>0.4</armorPenetrationSharp>
+							<armorPenetrationBlunt>0.33</armorPenetrationBlunt>
+						</li>
+						<li Class="CombatExtended.ToolCE">
+							<label>left claw</label>
+							<capacities>
+								<li>Scratch</li>
+							</capacities>
+							<power>6</power>
+							<cooldownTime>1.5</cooldownTime>
+							<linkedBodyPartsGroup>FrontLeftPaw</linkedBodyPartsGroup>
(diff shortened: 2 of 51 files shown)


Patches/AnimalCollabProj/AnimalCollabProj_CE_Patch_Races_Animal_Bison.xml
```diff
@@ -0,0 +1,41 @@
...
+				<xpath>Defs/ThingDef[defName="ACPBison"]/statBases</xpath>
+				<value>
+					<MeleeDodgeChance>0.08</MeleeDodgeChance>
+					<MeleeCritChance>0.18</MeleeCritChance>
+					<MeleeParryChance>0.15</MeleeParryChance>
+				</value>
+			</li>
+
+			<li Class="PatchOperationReplace">
+				<xpath>Defs/ThingDef[defName="ACPBison"]/tools</xpath>
```
N7Huntsman: The bison was patched under Races_Animal_Farm with RW 1.1 when VFA - Farm was added to vanilla. Probably better to just copy-paste those for consistency.

Patches/AnimalCollabProj/AnimalCollabProj_CE_Patch_Races_Animal_DuckBilledPlatypus.xml
```diff
@@ -0,0 +1,63 @@
...
+					<tools>
+						<li Class="CombatExtended.ToolCE">
+							<label>spurs</label>
+							<capacities>
+								<li>PlatypusSting</li>
+							</capacities>
+							<power>7</power>
+							<cooldownTime>1.9</cooldownTime>
+							<linkedBodyPartsGroup>Spurs</linkedBodyPartsGroup>
+							<armorPenetrationSharp>0.4</armorPenetrationSharp>
```
N7Huntsman: Might check the sharp AP here, since a lion bite only pierces 0.29mm.

Patches/AnimalCollabProj/AnimalCollabProj_CE_Patch_Races_Animal_Goose.xml
```diff
@@ -0,0 +1,62 @@
...
+		<success>Always</success>
+		<operations>
+
+			<li Class="CombatExtended.PatchOperationFindMod">
+				<modName>AnimalCollabProj</modName>
+			</li>
+
+			<li Class="PatchOperationAdd">
+				<xpath>Defs/ThingDef[@Name="GooseThingBase"]/statBases</xpath>
+				<value>
```
N7Huntsman: Goose was also added as a farm animal, just copy the stats.

Patches/AnimalCollabProj/AnimalCollabProj_CE_Patch_Races_Animal_Horse.xml
```diff
@@ -0,0 +1,96 @@
...
+			<li Class="CombatExtended.PatchOperationFindMod">
+				<modName>AnimalCollabProj</modName>
+			</li>
+
+			<li Class="PatchOperationAdd">
+				<xpath>Defs/ThingDef[
+					defName="ACPHorse" or 
+					defName="ACPWoolyHorse"
+				]/statBases</xpath>
+				<value>
```
N7Huntsman: Vanilla horse for reference. Given the differences in the various breeds of horses, your call if you think it's appropriate to adjust.
```
			<MoveSpeed>5</MoveSpeed>
			<MeleeDodgeChance>0.11</MeleeDodgeChance>
			<MeleeCritChance>0.30</MeleeCritChance>
			<MeleeParryChance>0.19</MeleeParryChance>
```

Patches/AnimalCollabProj/AnimalCollabProj_CE_Patch_Races_Animal_Silkie.xml
```diff
@@ -0,0 +1,51 @@
...
+				<xpath>Defs/ThingDef[defName="ACPSilkieChicken"]/statBases</xpath>
+				<value>
+					<MeleeDodgeChance>0.15</MeleeDodgeChance>
+					<MeleeCritChance>0</MeleeCritChance>
+					<MeleeParryChance>0</MeleeParryChance>
+				</value>
+			</li>
+
+			<li Class="PatchOperationReplace">
+				<xpath>Defs/ThingDef[defName="ACPSilkieChicken"]/tools</xpath>
```
N7Huntsman: No claw scratch attack like the vanilla chicken?

Patches/AnimalCollabProj/AnimalCollabProj_CE_Patch_Races_Animal_Bison.xml
sumghai: Good call.

Fixed in 51f311b

Patches/AnimalCollabProj/AnimalCollabProj_CE_Patch_Races_Animal_DuckBilledPlatypus.xml
sumghai: Whoops!

Fixed in 8fa8468

Patches/AnimalCollabProj/AnimalCollabProj_CE_Patch_Races_Animal_Goose.xml
sumghai: Ah, I see.

Fixed in ed9e97d

Patches/AnimalCollabProj/AnimalCollabProj_CE_Patch_Races_Animal_Horse.xml
sumghai: Probably not much difference between both breeds, so I'm setting them to vanilla values as well.

Fixed in 087aad7

Patches/AnimalCollabProj/AnimalCollabProj_CE_Patch_Races_Animal_Silkie.xml
sumghai: Yeah, just realized Silkies are otherwise identical to vanilla chickens.

Fixed in 46adb9e

Schatten5:
The missing animal sting hediffs are from the base mod, they have nothing to do with the patch.
I am getting this error every so often when guinea pigs are fighting human pawns:
RimWorldWin64_dGVp8JIC2g

sumghai:
@N7Huntsman @Schatten5 - Turns out there are **_two_** different versions of AnimalCollabProj 1.1:

* Xen's original version on Steam Workshop, which has missing hediff descriptions and includes ACPGuineaPig
* The GitHub version currently maintained by @cucumpear, which fixes the missing hediff descriptions, but removes ACPGuineaPig because Guinea Pigs are now included in the vanilla Core game.

The Steam version hasn't been updated to the same code as the GitHub version, and I'm not entirely sure how to make our patch work for both versions.

sumghai:
@Schatten5 - I did a bit more investigating, and it looks like ACPGuineaPig only shows up if you're running RimWorld 1.0, but not in RimWorld 1.1.

It's rather redundant work, but I can add a patch for ACPGuineaPig.

sumghai:
@Schatten5 - Your issue is "fixed" in 5f4bf05.

@N7Huntsman - Since the hediff description issue actually happens in vanilla, it's not really up to us to fix:

Annotation 2020-04-01 211901

As for the sting poison balancing, could you please point out the specific lines I need to fix? I can have a think about the actual values to be rebalanced then.

N7Huntsman:
Regarding the two different versions, the ideal thing to do would be to swap out `<li Class="CombatExtended.PatchOperationFindMod">`, which only accepts one mod, and reconfigure it for the vanilla PatchOperationFindMod, since it accepts a list of mod names

As for balancing the poisons, I think that'll require some additional patches. Find the damageDefs in the ACP files and it'll have some stuff to edit under 'additionalHediffs'. You can also find the hediffDef that the damages apply and tweak those as well.

sumghai:
> Regarding the two different versions, the ideal thing to do would be to swap out `<li Class="CombatExtended.PatchOperationFindMod">`, which only accepts one mod, and reconfigure it for the vanilla PatchOperationFindMod, since it accepts a list of mod names
Actually, both versions of AnimalCollabProj share the exact same mod name in their About.xml files - à la how both CE and CE:FT were named "Combat Extended".

(ASIDE: In the distant future, I may simply go through all of our existing patches and re-implement them using the vanilla `PatchOperationFindMod` for consistency, which has the added advantage of avoiding the `success=always` nonsense.)

> As for balancing the poisons, I think that'll require some additional patches. Find the damageDefs in the ACP files and it'll have some stuff to edit under 'additionalHediffs'. You can also find the hediffDef that the damages apply and tweak those as well.

Ah, fair enough. I'll take a look over the next few days.

sumghai:
@N7Huntsman - I hope I've managed to patched the venom hediffs properly to match the vanilla CE venom damages/hediffs. Essentially:

* `additionalHediffs` removed and replaced with `hediff` applied directly to the `DamageDef`s
* `DamageDefs` only harm the outer layer
* Hediffs now use `CombatExtended.HediffCompProperties_Venom`
* `VenomPerSeverity` adjusted such that platypus and scorpion stings will only build up to minor venom toxicity, before slowly fading 
  * As you mentioned, real life platypus and scorpion stings are usually not fatal, although the usual movement/conciousness temporary debuffs still apply

No errors when tested with both platypuses and scorpions from this mod.

N7Huntsman:
@sumghai Going to list my remarks/requests here, just to save me picking through all the files to add them as review comments:
- Panda body part armor is rather high at 0.59 RHA and 1.94 MPa, compared to other bears at 0.29 RHA and 0.94 MPa
- Megascorpion pincers deal Blunt damage. Is this intentional?
- Megascorpion sting attack may not be set to outside layers only, which is resulting in it damaging multiple body parts, each of which have a venom hediff applied to them, resulting in much faster poisoning.
- Megabadger armor and AP seem extremely low, given its description as being the "size of an elephant", I might reference the megafauna CE patch for similarly over-sized critters.
- The Xennagunna attack AP should probably be buffed a bit, given its size compared to the normal iguana.
- Both spirit wolves could probably have their armor and attack APs buffed. As semi-mythical creatures sort of like the Thrumbo, its fine for the armor to sort of look at the projectile sheet and decide "I want their armor to stop everything up to **this** kind of bullet**". Same for the dire wolves, but somewhat less so, in respect to their size and "power".
- While I understand they're intended to be fragile, the silkspider bite sharp AP seems a bit low. Thoughts?
-

sumghai:
@N7Huntsman - Just so you know, balancing animal and race stats is not my forte. The patches were often originally implemented by @AthenaSulisMinerva from various sources.

I would suggest that for future animal/race patches imported from FastTrack, we should try to get the basic PR approved quickly, and have someone else more experienced with animal race stats to do gameplay balances instead.

> Panda body part armor is rather high at 0.59 RHA and 1.94 MPa, compared to other bears at 0.29 RHA and 0.94 MPa

That's odd. None of the current patches specified any changes to panda or bear body part armors, nor could I find anything for vanilla CE bears

> Megascorpion pincers deal Blunt damage. Is this intentional?

Yes. The original mod also specified Blunt damage for Megascorpion pincers.

> Megascorpion sting attack may not be set to outside layers only, which is resulting in it damaging multiple body parts, each of which have a venom hediff applied to them, resulting in much faster poisoning.

Isn't it already set as such?

```
<li Class="PatchOperationAddModExtension">
	<xpath>Defs/DamageDef[
		defName="PlatypusSting" or
		defName="ScorpionSting"
	]</xpath>
	<value>
		<li Class="CombatExtended.DamageDefExtensionCE">
			<harmOnlyOutsideLayers>true</harmOnlyOutsideLayers>
		</li>
	</value>
</li>
```

> Megabadger armor and attack AP seem extremely low, given its description as being the "size of an elephant", I might reference the megafauna CE patch for similarly over-sized critters.

The Megafauna patches seldom specify armor ratings for similar creatures, and since the vanilla CE elephant has fairly low armor ratings specified, I'm going to use the ones for rhinoceroses instead.

Attack APs have been increased by a factor of 5.

Fixed in 27874fd 

> The Xennagunna attack AP should probably be buffed a bit, given its size compared to the normal iguana. 

Buffed in cbf44e8 

> Both spirit wolves could probably have their armor and attack APs buffed. As semi-mythical creatures sort of like the Thrumbo, for the armor it's fine to sort of look at the projectile sheet and decide "I want their armor to stop everything up to this kind of bullet**". Same for the dire wolves, but somewhat less so, in respect to their size and "power".

Cursory research suggests that the Spirit Wolves are from the _Okami_ video game, and were intended to have god-like powers, so I'm tempted to just use values similar to the vanilla Thrumbo.

Dire wolves were similarly buffed, with comparatively inferior stats.

Buffed in f227c77 

> While I understand they're intended to be fragile, the silkspider bite sharp AP seems a bit low. Thoughts?

As tiny insects engineered mainly to produce synthetic silk, I think it's reasonable they'd have fairly weak attack AP. No changes here.

> Tiger attack AP could use a buff. They're the largest big cat, but their AP is still on par with the far smaller jaguar.

I've buffed both the Lion and Tiger claw/bite attacks to be approx 1.5x times more powerful than the lesser big cats (e.g. panthers, jaguars).

Head attack blunt APs were only buffed 1.25x times, since I don't imagine their heads to be significantly larger than other big cats, while a larger body mass would definitely result in greater momentum for a head-butt attack.

Buffed in 1c1aefd

N7Huntsman:
> Just so you know, balancing animal and race stats is not my forte. ... I would suggest that for future animal/race patches imported from FastTrack, we should try to get the basic PR approved quickly, and have someone else more experienced with animal race stats to do gameplay balances instead.
Fair enough--it's one of those stranger sort of grey areas where you sort of have to use the spreadsheet as a baseline and then sort of "feel" your way through. Probably wouldn't be good practice to have it merged and then go through for balance (if that's what you were suggesting?) just because inevitably something would get merged and the balance pass would be forgotten about, but I'd be onboard with having something PR/send files with balance changes to your branch.

PR #1163, opened 2020-03-30
Fixes to android tiers patch and support for TX series

Tostov:
## Additions

Added support for the TX and TXK series added by _Android tiers - TX Series_, which were previously part of the Android Tiers++ patch (which is now the TX series mod patchwise, as AT++ was intergrated barring the TX series, and the TX androids were the only thing that needed patching in the first place).

## Changes

Fixed missing mod requirements in PatchOperationFindMod operations for two patches that were missing the new one for 1.1.

## Testing


N7Huntsman (approved):
Looks good.

PR #1169, opened 2020-04-01
Update to (Rim) Contractor's Arsenal

Tostov:
## Additions

Patches for new weapons added by the base mod.
Patches added for new pawn types added by the base mod.
Adjusted mod find patch to allow for either regular or factionless variant of base mod
The rocket launcher in the base mod is not single use, but has been converted into so for CE.

## Reasoning

In general weapons have been matched up to their respective closest match for _advanced_ calibers (excluding the RSH12, which has been downscaled to 8x35mm). The AT45 has been given a irregular HEAT/EMP effect since it's supposed to be a charge rocket launcher thing used primarly for anti-mechanoid work, which this change makes it highly effective at.

The survival rifle's description suggests that it doesn't need ammuntion to function, and as such it doesn't, though it uses the effects of a smaller than expected gauge (6x24 vs 8x35), has significantly reduced rate of fire and increased weapon cooldown, in addtion to retaining the advanced component cost that is usually eschewed in CE charge weapons.

## Testing

### DIFF
--- a/Patches/Rim Contractors Arsenal/ThingDefs_Misc/MeleeContractors.xml
+++ b/Patches/Rim Contractors Arsenal/ThingDefs_Misc/MeleeContractors.xml
@@ -1,0 +1,80 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Patch>
+  <Operation Class="PatchOperationSequence">
+	<operations>
+	  <li Class="PatchOperationFindMod">
+			
+		<mods>
+		<li>Contractors Arsenal</li>
+		<li>Contractors Arsenal (Factionless)</li>
+		</mods>
+			
+		<match Class="PatchOperationSequence">
+		<operations>
+
+			<li Class="PatchOperationReplace">
+				<xpath>Defs/ThingDef[defName="Melee_ContractorsCombatKnife"]/tools</xpath>
+				<value>
+					<tools>
+						<li Class="CombatExtended.ToolCE">
+							<label>handle</label>
+							<capacities>
+								<li>Poke</li>
+							</capacities>
+							<power>1</power>
+							<cooldownTime>1.26</cooldownTime>
+							<armorPenetrationBlunt>0.250</armorPenetrationBlunt>
+							<linkedBodyPartsGroup>Handle</linkedBodyPartsGroup>
+						</li>
+						<li Class="CombatExtended.ToolCE">
+							<label>blade</label>
+							<capacities>
+								<li>Cut</li>
+							</capacities>
+							<power>10</power>
+							<cooldownTime>1.18</cooldownTime>
+							<armorPenetrationBlunt>0.36</armorPenetrationBlunt>
+							<armorPenetrationSharp>7.32</armorPenetrationSharp>
+							<linkedBodyPartsGroup>Blade</linkedBodyPartsGroup>
+						</li>
+						<li Class="CombatExtended.ToolCE">
+							<label>point</label>
+							<capacities>
+								<li>Stab</li>
+							</capacities>
+							<power>11</power>
+							<cooldownTime>1.26</cooldownTime>
+							<chanceFactor>1.33</chanceFactor>
+							<armorPenetrationBlunt>0.25</armorPenetrationBlunt>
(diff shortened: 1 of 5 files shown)


Patches/Rim Contractors Arsenal/ThingDefs_Misc/MeleeContractors.xml
```diff
@@ -0,0 +1,80 @@
...
+						</li>
+						<li Class="CombatExtended.ToolCE">
+							<label>blade</label>
+							<capacities>
+								<li>Cut</li>
+							</capacities>
+							<power>10</power>
+							<cooldownTime>1.18</cooldownTime>
+							<armorPenetrationBlunt>0.36</armorPenetrationBlunt>
+							<armorPenetrationSharp>7.32</armorPenetrationSharp>
```
N7Huntsman, line 37: This strikes me as being quite high, especially given that the stab attack doesn't penetrate significantly more.

Patches/Rim Contractors Arsenal/ThingDefs_Misc/RangedContractors.xml
```diff
@@ -380,9 +411,435 @@
...
+				<FireModes>
+					<aimedBurstShotCount>2</aimedBurstShotCount>
+					<aiUseBurstMode>FALSE</aiUseBurstMode>
+					<aiAimMode>Snapshot</aiAimMode>
+				</FireModes>
+			</li>
+			
+			<!--SCR Survival Rifle-->
+			
+			<li Class="CombatExtended.PatchOperationMakeGunCECompatible">
```
N7Huntsman, line 673: I can't say I'm a big fan of giving the weapon infinite ammo. Perhaps you could just make it accept a variety of different ammo types, instead? (which would be accomplished by creating a custom ammoset for it)

N7Huntsman (changes requested):

Patches/Rim Contractors Arsenal/ThingDefs_Misc/MeleeContractors.xml
Tostov: Just like the power sword, the knife pretty much just adds 7mm of sharp penetration onto the weapon's specs.

Patches/Rim Contractors Arsenal/ThingDefs_Misc/RangedContractors.xml
Tostov: Ultimately, this is once of those instances where I feel that if we had a recharging ammo pool type thing, It'd be ideal.

That being said, the issue with patching in a new ammo set is that it could:
A) Just use a single bullet type
In which instance it feels kind of silly. The idea of the weapon is to pretty much cut off the need for ammo supply. The description itself states that it needs no ammo to function. But that being said, if it did use one ammo type, it'd run into the issue where noone would have the appropriate ammo to run it it, since you'd be looking at the average thing having a myriad of possible enemies carrying different ammo. It'd make it no different than any other charge rifle, just needing one specific type of regular ammo rather than charge bullets, which undermines the general survival aspect of the weapon.
B)Use multiple bullet types
The issue here is that it either has to have a metric ton of side patches to check for something that uses that ammo (which might not even be possible via XML), or it'd force every possible ammo it could use to spawn soley for its own usage, even if they couldn't be used in any other gun, which would severly bloat the size of the ammo list even if the player isn't using a large variety of mods.

Ulitimately, I feel that giving one or two guns the ability to fire an infinite amount of ammo is fine. By building this instead of a conventional charge weapon, you're sacrificing a serious spread of stats in exchange for general ammo logevity, but if you can make charge weapons in the first place, aquiring ammo for them will almost certainly not be a serious issue.

I would consider tweaking the weapon tags/recipe/cost so that it is difficult to aquire without constructing it yourself, if you think that would resolve the issue, perhaps?

Patches/Rim Contractors Arsenal/ThingDefs_Misc/MeleeContractors.xml
N7Huntsman: Fair enough.

N7Huntsman (approved):

ghost (approved):

PR #1185, opened 2020-04-04
Upload patchfix for kijin

Tostov:
## Changes

Fixed bodies patch findmod operation to be consistent with other files patch operations.

## Testing

### DIFF
--- a/Patches/Kijin 2.0/Bodies/Mod_Bodies_Kijin.xml
+++ b/Patches/Kijin 2.0/Bodies/Mod_Bodies_Kijin.xml
@@ -7,7 +7,7 @@
 			
 			<li Class="PatchOperationFindMod">
 				<mods>
-					<li>Kijin Race</li>
+					<li>Kijin Race 2.0</li>
 				</mods>
 				<match Class="PatchOperationSequence">
 					<operations>
--- a/Patches/Kijin 2.0/ThingDef_Misc/Kijin_Apparel.xml
+++ b/Patches/Kijin 2.0/ThingDef_Misc/Kijin_Apparel.xml
@@ -82,6 +82,14 @@
 				</value>
 			</li>
 			
+			<li Class="PatchOperationAdd">
+				<xpath>Defs/ThingDef[defName="Kijin_EpicArmor"]/apparel/bodyPartGroups</xpath>
+				<value>
+					<li>Hands</li>
+					<li>Feet</li>
+				</value>
+			</li>
+			
 			<li Class="PatchOperationReplace">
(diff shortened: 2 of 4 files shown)


N7Huntsman (approved):

N7Huntsman (approved):

N7Huntsman:
Currently getting `Config error in Kijin_SmokeBomb: CompExplosiveCE requires Normal ticker type.` on start up. Likely just related to the recent changes made to the explosive comps.

I'm unfamiliar with the base Kijin mod, but from a strictly numerical standpoint, their armor seems somewhat high (not helped by the fact it scales by 150% due to their base health scale). If that's on par compared to their performance in vanilla, then so be it, I suppose. However, please double-check that it's being applied to body parts like you want it to--while testing, I noticed the Kijin's hand and foot were injured by a wolf's paw scratch, that I don't think was supposed to have high enough sharp AP to cause a scratch injury.

Their melee dealing stun damage or igniting the target is also pretty broken. Rather than having them guaranteed by making them the Damage def, would it be practical and less broken to have the stun/flame damage do something like the Royalty weapons? This is from a weapon and I haven't tested this on a pawn melee capacity, but it may be worth trying?
```
				<li Class="CombatExtended.ToolCE">
					<label>edge</label>
					<capacities>
						<li>Cut</li>
					</capacities>
					<power>40</power>
					<extraMeleeDamages>
					<li>
						<def>Flame</def>
						<amount>5</amount>
						<chance>0.3</chance>
					</li>
					</extraMeleeDamages>					
					<cooldownTime>2.06</cooldownTime>
					<chanceFactor>0.30</chanceFactor>
					<armorPenetrationBlunt>3.20</armorPenetrationBlunt>
					<armorPenetrationSharp>16</armorPenetrationSharp>
					<linkedBodyPartsGroup>Edge</linkedBodyPartsGroup>
				</li>
```

N7Huntsman (changes requested):
Review comment. I know it's outside the scope of what you actually did for this patch at present, but it is something that should be addressed.

Tostov:
Ahh sorry, I did run a test with it an it seemed pretty good. I'll PR the changes in a day or so, just been somewhat busy the last few days or so.

I'll add hand a feet coverage to the armor too. I feel the value is fine, it's a fairly expensive set and the vanilla protective value is the same as power armor, so scaling it down to 14mm hits a nice medium of not making power armor irrelavant while still keeping it at a reasonably good spot. Since it has a inbuilt shield anyway, I'm pretty sure it can't be used in conjunction with ranged weapons ever either.

N7Huntsman:
@Tostov To clarify, I was talking about their natural armor, rather than apparel.

N7Huntsman (approved):

ghost (approved):

PR #1186, opened 2020-04-04
Patch update for Dragonian Race 1.1

Tostov:
## Additions

Added patch for new armored dress, helm and weapon.

## Changes

Gave the wool a small amount of electric resist.

## Testing

### DIFF
--- a/Patches/Dragonian Race/ThingDefs_Misc/DR_Apparel.xml
+++ b/Patches/Dragonian Race/ThingDefs_Misc/DR_Apparel.xml
@@ -93,6 +93,14 @@
 				</value>
 			</li>
 			
+			<li Class="PatchOperationAdd">
+				<xpath>Defs/ThingDef[defName="DR_BattleDress"]/apparel/bodyPartGroups</xpath>
+				<value>
+					<li>Hands</li>
+					<li>Feet</li>
+				</value>
+			</li>
+			
 			<li Class="PatchOperationReplace">
 				<xpath>Defs/ThingDef[defName="DR_BattleHelm"
 				]/statBases/StuffEffectMultiplierArmor</xpath>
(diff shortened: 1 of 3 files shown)


Patches/Dragonian Race/ThingDefs_Misc/DR_Apparel.xml
```diff
@@ -65,6 +65,60 @@
 				</value>
 			</li>
 			
+			<!-- ========== Armor ========== -->
+			
+			<li Class="PatchOperationReplace">
```
N7Huntsman, line 70: You've confirmed that the combination of material armor + statBases armor produces the range of armor values you want?

N7Huntsman (changes requested):

Patches/Dragonian Race/ThingDefs_Misc/DR_Weapon.xml
```diff
@@ -0,0 +1,96 @@
...
+				<xpath>Defs/ThingDef[defName="DR_BattleHammer"]/tools</xpath>
+				<value>
+					<tools>
+						<li Class="CombatExtended.ToolCE">
+							<label>handle</label>
+							<capacities>
+								<li>Poke</li>
+							</capacities>
+							<power>5</power>
+							<extraMeleeDamages>
```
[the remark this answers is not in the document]
Tostov: The original weapon in this instance addtionally zaps people with the handle poke as well, rather than just with the head like with the Zeushammer.

Given the way secondary damage works, it is admittedly largely redundant since the majority of the time it won't matter (since secondary damage doesn't seem to go off if the main damage gets blocked).

Patches/Dragonian Race/ThingDefs_Misc/DR_Apparel.xml
Tostov: Yes.

N7Huntsman (approved):

ghost (approved):

PR #1193, opened 2020-04-06
More practical traits 1.1 Fix

Tostov:
## Additions

CE adjustments for Swordsman of Squin Trait

## Changes

Fixed the trait patches that were causing errors.

## Testing

### DIFF
--- a/Patches/More Practical Traits/TraitDefs/Traits_MPT.xml
+++ b/Patches/More Practical Traits/TraitDefs/Traits_MPT.xml
@@ -16,14 +16,14 @@
 			<li Class="PatchOperationReplace">
 				<xpath>Defs/TraitDef[defName="VanyaTA01mhgg"]/degreeDatas/li/statOffsets/ArmorRating_Sharp</xpath>
 				<value>
-					<ArmorRating_Sharp>18</ArmorRating_Sharp>
+					<ArmorRating_Sharp>12</ArmorRating_Sharp>
 				</value>
 			</li>
 			
 			<li Class="PatchOperationReplace">
 				<xpath>Defs/TraitDef[defName="VanyaTA01mhgg"]/degreeDatas/li/statOffsets/ArmorRating_Blunt</xpath>
 				<value>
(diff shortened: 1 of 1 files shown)


N7Huntsman:
Man, some of these traits are bizarre... At any rate, there are a few that involve armor, combat skills, or other CE-related stuff that were overlooked.
- VanyaTC01rrwg "Delicate" : Blunt Armor, frankly might replace this with an increased `IncomingDamageFactor` if it makes sense--not sure why this trait gets extra blunt armor)
- I _think_ carrying capacity is no longer used with CE loaded. Should the patchops that adjust traits with it be a Replace instead of an Add? Maybe it's harmless, though. I'm not certain.
- PTA14yg "Brave" : Do we think it might be necessary to change the label to "valiant" or "courageous" or something, for the sake of differentiating it from the CE "Brave" trait?
- PTB02tb "Impregnable" : Armor values
- LingTB01lth "lost quality" : Weirdly, this trait sets mass to 0. Probably worth double check that it doesn't cause issues with any kind of calculations CE uses for hit box or something.
- PTC13cjz "punisher" : armor values
- PTF01hy "red moon" : ShootingAccuacyPawn might need buffed here?
- "Moe", "warrior of the sun" , "doomsday boxer" and various others add pretty high armor values. Are they intended to be this high?

N7Huntsman (changes requested):
Review comment.

Tostov:
Given this is one of those instances where it may be a reference to something rather than a specific description of the person, I may just scale up the blunt armor stat anyway.

Speaking of which yeah, some of the traits add hilarously high amounts of armor for vanilla. Show moe for example is just a flat 150% buff to all armor values (it makes the pawn otherwise incapable of everything outside of literally being a meatshield though). I'll scale the specs down to 8mm/12MPa at equivalent 1.0 though, since that should keep them roughly consistent without making unkillable nudists, but overall most of the traits that have the absurd armor values tend to fall outside of the balanced side of the curve for MPT anyway.

I will get to fixing the other armor stats/adding accuracy stuff to red moon.

PR #1198, opened 2020-04-07
Vanilla Weapons Expanded Fixes

ghost:
## Changes

- Marksman rifle now uses the correct ammo
- Service Rifle now properly patched
- Whip ranged damage down from **14** to **10**

## Reasoning
- Whip felt a little bit too strong, my reasoning is that it should take you a lot more to kill / down someone with a whip, still open for discussion.

## Testing

- [x] (For compatibility patches) ... with and without patched mod loaded
- [x] Playtested a colony (1h)


N7Huntsman:
If we're still concerned about the whip being too strong, perhaps it would be appropriate to create a RangedCut damageDef to substitute for the RangeStab? The whip is sort of in an odd place in that it makes sense for it to deal sharp damage, but also has extremely little sharp armor penetration.

Also, I'm somewhat unclear on what you changed with the sawed-off and SMG in this PR.

onigamii:
> If we're still concerned about the whip being too strong, perhaps it would be appropriate to create a RangedCut damageDef to substitute for the RangeStab? The whip is sort of in an odd place in that it makes sense for it to deal sharp damage, but also has extremely little sharp armor penetration.
> 
> Also, I'm somewhat unclear on what you changed with the sawed-off and SMG in this PR.

Perhaps may be the best thing. As for the sawed-off and the smg i just moved it from where you initially put it to after smg's call it ocd

onigamii:
### Additions
- Added a new damage type Ranged Cut for the whip (My first time messing with this damage type so feel free to improve it)

### Changes
- Reverts the Fire Extinguisher research prerequisite to Firefoam from Launchers
- Lowered the Whip damage to 8 from 10 (Still felt a bit too strong, open for discussion)

PR #1202, opened 2020-04-08
Vanilla Factions Expanded - Settlers

ghost:
## Additions
Patches the Settlers faction addon from Vanilla Factions Expanded

## Testing
- [x] Playtested a colony (specify how long)

### DIFF
--- a/Patches/Core/ThingDefs_Misc/Apparel_Belts.xml
+++ b/Patches/Core/ThingDefs_Misc/Apparel_Belts.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <Patch>
 
-	<!-- ========== Simple Helmet ========== -->
+	<!-- ========== Shield belt ========== -->
 
 	<Operation Class="PatchOperationAdd">
 		<xpath>Defs/ThingDef[defName="Apparel_ShieldBelt"]/statBases</xpath>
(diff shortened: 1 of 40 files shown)


onigamii:
Alright after some testing it's done, don't think i need to change anything else.

N7Huntsman (dismissed):
Looks good. Weapons and chemshine work fine. @onigamii Gatling gun rate of fire feels a bit high for the typical operator, maybe? Mag size is probably kinda high as well. 

For the purposes of finding real-world specs to use, we'll consider this an 1877 "Bulldog" Gatling gun.

onigamii:
@N7Huntsman Alright, been looking around and agree that the mag size is quite high so based on what i read on wikipedia i'll reduce the mag size to 40 - 50 maybe?. And for the rate of fire i keep seeing around 800 rpm for the 5 barreled version.

So here's some ideas:
**Mag size** 100 > 50 / 40
**Rate of Fire** 5 > 7

N7Huntsman (dismissed):

MMMiracles (dismissed):
I played about 6-7 hours on Diego Dire with the bandit start. No real balance complaints past the gatling turret, which got changed to something more reasonable. It didn't seem like the dynamite sticks had the randomization on their detention timer like advertised but honestly I'd avoid using a weapon that could explode in my face by sheer chance, that and the AI probably doesn't need any more handicaps as is if they happen to spawn with them.

ghost:
Alright, made some changes that i believe are the last.

**Musket** 
- If Run and Gun is loaded, disables it from firing while running.

**Dynamite** 
- Now uses ` Woody ` category to allow use of wood added by other mods

PR #1209, opened 2020-04-09
(bugfix) Fix bug where turrets cannot be loaded

### LINKED ISSUE #1188: Turrets cannot be reloaded // Turrets reloaded with wrong ammo type
Hello,

there is a bug when it is not possible to reload the turrets or when the turrets being reloaded with wrong ammo type.

Second part is easy to reproduce:
[Images](https://imgur.com/a/Pz6sSG6)
1. Spawn ammo not in stacks, not the default ammo type (e.g. AP-I)
2. Place turrets. Everything is OK: no ammo-switch and reload occur
2. Designate a stockpile. Wait until some ammo will be merged
3. As soon as there more than 100 ammo in 1 stack, the auto-switch of ammo occur and turret is being reloaded. Error message occur as well.

The first part (unable to reload the turrets) is harder to reproduce. To do it, I have to recreate several maps until the effect occur. In my normal game (with a lots of mods) this occur always.
Here is a [video demonstration](https://youtu.be/E7gou_vf9_g) without any additional mods.

[Steam discussion](https://steamcommunity.com/workshop/filedetails/discussion/1631756268/2145343824297003752/)

MadaraUchiha:
- Also increase ammo search range from 40 tiles to 100 tiles.
- Fix bug where lastSurrdoundingAmmoCheck would not be updated if
GenClosestAmmo cache hit.
- Fixes #1188

## Additions

N/A

## Changes

Adjustments to other features made in this merge, e.g.
- Increase turret ammo search radius from 40 tiles to 100 tiles.
- Fix bug where `GenClosestAmmo` cache hit would cause the TTL to never be updated.

## References

- Closes #1188 

## Reasoning

- Turrets are generally found on the edge of your base, far away from your stockpiles. Hence, the increase in search radius. It may be prudent to add a reason to why a turret can't be reloaded next time (something along the lines of "No compatible ammo in range")

## Alternatives

Remove the `GenClosestAmmo` class, as suggested by @NoImageAvailable.

## Testing

- [x] Compiles without warnings
- [ ] (For compatibility patches) ...with and without patched mod loaded
- [x] Playtested a colony (1 hour, though not very extensively)


NoImageAvailable:
What's the reasoning behind the try-catch block?

N7Huntsman:
We certainly appreciate the help, though unfortunately the issue persists. In testing, I'm seeing some truly bizarre behavior, but I'm hopeful you'll find the details helpful. I've included video of my testing, and a trace for the reload job error I encountered.
Trace: [https://pastebin.com/Q7AUjSvS](url)

NoImageAvailable:
All those issues sound like issues with the caching system. Unless there is some 1.1 specific reason it has to be there it should be rolled back.

MadaraUchiha:
> 
> 
> What's the reasoning behind the try-catch block?

The variable tracking the last check should be updated regardless of whether the cache lookup bails early. I could have added the last check update before every return, but figured this was easier and clearer.

I do agree that if there's no real good reason for this cache class, it should be removed.

NoImageAvailable:
So here's the case for reverting the cache class
Con:
- It saves some performance on colonies with lots of turrets
- Simply reverting it will wipe out Mechanoid reloading along with it

Pro:
- It's extremely buggy and will likely take extended effort to make work
- Turret reloading isn't a particularly huge performance drain to begin with
- It introduces lag in reloads even if it works

@MadaraUchiha can you ditch the caching in favor of 1.0 style reloading while keeping mech functionality intact (potentially by reimplementing it in a more reasonable format such as a ThinkNode)?

MadaraUchiha:
My RimWorld specific C# experience is very much lacking, but I'll give it my best shot. If you have any resources that you thing can be useful (i.e. save me on reading decompiled RimWorld code) please let me know :).

N7Huntsman:
I'll ping @zhrocks11 on this for the sake of bringing him into the loop and putting him in contact with @MadaraUchiha . I believe Zhrocks is taking a somewhat similar approach of rolling back some stuff, so hopefully you two can communicate and avoid duplicating work.

